PR #126, opened 2018-04-26
Webview interaction and system webview [in progress]

unpluggedk:


IdentityCore/src/MSIDError.h
```diff
@@ -69,9 +69,49 @@ typedef NS_ENUM(NSInteger, MSIDErrorCode)
...
+    /*!
+     The state returned by the server does not match the state that was sent to
+     the server at the beginning of the authorization attempt.
+     */
+    MSALErrorInvalidState = -52501,
+    /*!
+     Interaction required errors occur because of a wide variety of errors
+     returned by the authentication service.
+     */
+    MSIDErrorMismatchedUser             = -52101,
```
oldalton: just curious, when is this error returned?

IdentityCore/src/MSIDError.h
```diff
@@ -69,9 +69,49 @@ typedef NS_ENUM(NSInteger, MSIDErrorCode)
...
+    MSIDErrorNoAuthorizationResponse    = -52102,
```
oldalton: nit: no authorization code?

IdentityCore/src/MSIDError.h
```diff
@@ -69,9 +69,49 @@ typedef NS_ENUM(NSInteger, MSIDErrorCode)
...
+    /*!
+     An interactive authentication session is already running with the
+     SafariViewController visible. Another authentication session can not be
+     launched yet.
+     */
+    MSIDErrorInteractiveSessionAlreadyRunning = -51021,
+    /*!
+     An interactive authentication session failed to start.
+     */
+    MSIDErrorInteractiveSessionStartFailure = -51022,
```
oldalton: what is the case when this would happen?

IdentityCore/src/MSIDError.m
```diff
@@ -49,3 +49,20 @@
 }
 
 
+MSIDErrorCode MSIDErrorCodeForOAuthError(NSString *oauthError, MSIDErrorCode defaultCode)
+{
+    if (oauthError && [oauthError caseInsensitiveCompare:@"invalid_request"] == NSOrderedSame)
```
oldalton: We already have this code in token response I think. Can we combine it?

IdentityCore/src/oauth2/MSIDRequestParameters.h
```diff
@@ -40,5 +40,7 @@
 
 @property (readwrite) NSString *loginHint;
 @property (readwrite) NSString *extraQueryParameters;
+@property (readwrite) NSString *promptBehavior;
```
oldalton: I think prompt behavior was something AAD V1 specific?

IdentityCore/src/oauth2/aad_v1/MSIDAADV1Oauth2Factory.m
```diff
@@ -151,15 +152,70 @@ - (MSIDLegacySingleResourceToken *)legacyTokenFromResponse:(MSIDTokenResponse *)
...
+    
+    // if value is nil, it won't appear in the dictionary
+    NSMutableDictionary *queryParams = [NSMutableDictionary dictionaryWithObjectsAndKeys:
+                                        MSID_OAUTH2_CODE, MSID_OAUTH2_RESPONSE_TYPE,
+                                        [requestParams clientId], MSID_OAUTH2_CLIENT_ID,
+                                        [requestParams resource], MSID_OAUTH2_RESOURCE,
+                                        [requestParams redirectUri], MSID_OAUTH2_REDIRECT_URI,
+                                        state, MSID_OAUTH2_STATE,
+                                        requestParams.promptBehavior, @"prompt",
+                                        @"1", @"hashchrome", //to hide back button in UI
```
oldalton: just curious, do we still need this?

IdentityCore/src/util/MSIDCryptoHelper.h
```diff
@@ -27,8 +27,8 @@
 
 #import <Foundation/Foundation.h>
 
-@interface MSIDWebviewRequest : NSObject
+@interface MSIDCryptoHelper : NSObject
 
-@property NSURL *startURL;
++ (NSData *)msidSHA256fromString:(NSString *)string;
```
oldalton: Don't we already have this functionality in some category (e.g. msidComputeSHA256)?

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,33 +26,114 @@
...
 }
 
-+ (id<MSIDWebviewInteracting>)systemWebviewControllerWithRequestParameters:(MSIDRequestParameters *)parameters
-                                                                   factory:(MSIDOAuth2Factory *)factory;
++ (void)startWebviewAuth:(id<MSIDWebviewInteracting>)webviewController
+                 context:(id<MSIDRequestContext>)context
+       completionHandler:(MSIDWebUICompletionHandler)completionHandler
 {
-    return nil;
+    if (![self setCurrentWebSession:webviewController])
```
oldalton: nit: should we check this before creating a controller?

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+    if ([NSString msidIsStringNilOrBlank:url.absoluteString])
+    {
+        if (error){
+            *error = MSIDCreateError(MSIDOAuthErrorDomain, MSIDErrorNoAuthorizationResponse, @"No authorization response received from server.", nil, nil, nil, context.correlationId, nil);
+            return nil;
+        }
+    }
+    
+    // Check for WPJ response
+    if([url.absoluteString hasPrefix:@"msauth://"])
```
oldalton: nit: space

oldalton: I think "msauth" is something really AAD specific, so let's discuss if we should keep it here.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+    {
+        NSString *query = [url query];
+        NSDictionary *queryParams = [NSDictionary msidURLFormDecode:query];
+        NSString *appURLString = [queryParams objectForKey:@"app_link"];
+        
+        MSIDWebWPJAuthResponse *response = [MSIDWebWPJAuthResponse new];
```
oldalton: Let's discuss, if MSIDWebOAuth2Response should create its subclass.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+    if (!stateVerified)
+    {
+        if (error){
+            *error = MSIDCreateError(MSIDOAuthErrorDomain, MSALErrorInvalidState, @"State returned from the server does not match", nil, nil, nil, context.correlationId, nil);
+            return nil;
+        }
+    }
+    
+    NSString *code = parameters[MSID_OAUTH2_CODE];
+    NSString *cloudHostName = parameters[MSID_AUTH_CLOUD_INSTANCE_HOST_NAME];
```
oldalton: cloud host name is something AAD specific, let's discuss if it should be here

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+    response.url = url;
+    response.oauthError = oauthError;
+    
+    return response;
+    
+}
+
++ (NSError *)oauthErrorFromDictionary:(NSDictionary *)dictionary
+{
+    NSUUID *correlationId = [dictionary objectForKey:MSID_OAUTH2_CORRELATION_ID_RESPONSE] ?
```
oldalton: correlation ID is something AAD specific

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+    [[NSUUID alloc] initWithUUIDString:[dictionary objectForKey:MSID_OAUTH2_CORRELATION_ID_RESPONSE]]:nil;
+    
+    NSString *serverOAuth2Error = [dictionary objectForKey:MSID_OAUTH2_ERROR];
+    
+    if (serverOAuth2Error)
+    {
+        NSString *errorDescription = dictionary[MSID_OAUTH2_ERROR_DESCRIPTION];
+        NSString *subError = dictionary[MSID_OAUTH2_CORRELATION_ID_RESPONSE];
```
oldalton: I think subError shouldn't come from MSID_OAUTH2_CORRELATION_ID_RESPONSE :)

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,127 @@
...
+
+// TODO: check if we have it in MSAL
++ (BOOL)verifyStateFromDictionary: (NSDictionary*) dictionary
+                          context:(id<MSIDRequestContext>)context
+{
+    NSDictionary *state = [NSDictionary msidURLFormDecode:[[dictionary objectForKey:MSID_OAUTH2_STATE] msidBase64UrlDecode]];
+    if (state.count != 0)
+    {
+        NSString *authorizationServer = [state objectForKey:@"a"];
+        NSString *resource            = [state objectForKey:@"r"];
```
oldalton: This looks very AAD v1 specific

IdentityCore/src/webview/systemWebview/MSIDSFAuthenticationSession.h
```diff
@@ -25,13 +25,19 @@
...
+#import "MSIDSystemWebviewController.h"
 
-@implementation MSIDSystemWebviewRequest
+@interface MSIDSFAuthenticationSession : NSObject
 
-- (id)initWithRequestParameters:(MSIDRequestParameters *)requestParameter
-{
-    return nil;
-}
+- (id)initWithURL:(NSURL *)url
```
oldalton: nit: id -> instancetype

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.h
```diff
@@ -26,15 +26,16 @@
...
 
 #import <Foundation/Foundation.h>
-#import "MSIDWebviewRequest.h"
+#import "MSIDSystemWebviewController.h"
 
-@class MSIDRequestParameters;
+@interface MSIDSafariViewController : NSObject
 
-@interface MSIDEmbeddedWebviewRequest : MSIDWebviewRequest
+- (id)initWithURL:(NSURL *)url
```
oldalton: nit: id -> instancetype

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,130 @@
...
+- (id)initWithURL:(NSURL *)url
+          context:(id<MSIDRequestContext>)context
+{
+    self = [super init];
+    if (self)
+    {
+        _url = url;
+        _context = context;
+        
+        _safariViewController = [[SFSafariViewController alloc] initWithURL:url entersReaderIfAvailable:NO];
```
oldalton: how do we ensure that we're on the main thread?

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.h
```diff
@@ -26,15 +26,16 @@
...
+          context:(id<MSIDRequestContext>)context;
 
-- (id)initWithRequestParameters:(MSIDRequestParameters *)requestParameter;
+- (BOOL)handleURLResponse:(NSURL *)url;
+- (BOOL)start;
 
-@property NSURL *endURL;
-@property NSDictionary *headers;
+@property id<MSIDSystemWebviewResponseDelegate> webviewDelegate;
```
oldalton: where is webviewDelegate set? Also, if it will be set in a "wrapper" class, who has reference to this controller, this should be weak.

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
```diff
@@ -26,19 +26,91 @@
...
+#import "MSIDSFAuthenticationSession.h"
+#import "MSIDSafariViewController.h"
 
 @implementation MSIDSystemWebviewController
+{
+    id<MSIDRequestContext> _context;
+    MSIDWebUICompletionHandler _completionHandler;
+
+#ifdef __IPHONE_11_0
+    MSIDSFAuthenticationSession *_authSession;
```
oldalton: nit: this is not set anywhere

IdentityCore/src/MSIDError.h
unpluggedk: no, it's just url nil response.

IdentityCore/src/MSIDError.h
unpluggedk: SFAuthenticationSession: when start is called on a cancelled session
SafariViewController, when we can't find the current viewcontroller to present from.

IdentityCore/src/MSIDError.m
unpluggedk: Where in token response do we have this?

IdentityCore/src/oauth2/aad_v1/MSIDAADV1Oauth2Factory.m
unpluggedk: This code is JZ's code. I was merely copying and pasting for now to make it compile.
Will remove this code for now from mine.

IdentityCore/src/oauth2/MSIDRequestParameters.h
unpluggedk: It also exists for MSAL in a different value.

        case MSALForceLogin : return @{ @"prompt" : @"login" };
        case MSALForceConsent : return @{ @"prompt" : @"consent" };
        case MSALSelectAccount : return @{ @"prompt" : @"select_account" };

IdentityCore/src/webview/MSIDWebviewAuthorization.m
unpluggedk: Thought about that too, it seems just a little more logical this way although it might be less efficient.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: I'm with you, I'd rather have this in a initializer for MSIDWebAADAuthResponse and check for nil there.
@jasoncoolmax what do you think?

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: Like above, I'd like to see it moved.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: Like above :)

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: MSAL also has this unless I am mistaken

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: No it shouldn't :p

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: Yes, the state handling is also different.
Was manually merging JZ's code, this need to be changed;

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
unpluggedk: not a nit, huge mistake :p this should've been set 👍

IdentityCore/src/MSIDError.h
unpluggedk: When parameter sets a user, and we force webview to login as a different user. 
We handled it as an error in MSAL,  we can discuss about this.

IdentityCore/src/MSIDError.m
unpluggedk: done, i've changed TokenResponse to call this.

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
unpluggedk: Good point, I think it is logical that we simply check if we are on main thread when start is called, and return NO if we aren't.

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,33 +26,213 @@
...
+    [self startWebviewAuth:systemWebviewController
+                   context:context
+         completionHandler:completionHandler];
+}
+
++ (void)startWebviewAuth:(id<MSIDWebviewInteracting>)webviewController
+                 context:(id<MSIDRequestContext>)context
+       completionHandler:(MSIDWebUICompletionHandler)completionHandler
+{
+    if (![self setCurrentWebSession:webviewController])
```
jasoncoolmax: seems we are not able to `setCurrentWebSession` for the second time, as there is no code to release `s_currentWebSession` when a webview call is completed? :)

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,33 +26,213 @@
...
+        completionHandler(nil, error);
+        return;
+    }
+    
+    if (![s_currentWebSession start])
+    {
+        NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractiveSessionStartFailure, @"Interactive web session failed to start.", nil, nil, nil, context.correlationId, nil);
+        
+        completionHandler(nil, error);
+    }
```
jasoncoolmax: Since we have already passed `completionHandler` to system/embedded webview, shall we let the webview class to call the completion block if it fails to start?
The `start` action is kind of async because it involves UI/main thread, it may not be able to return a value immediately to indicate if the start action succeeds :)

jasoncoolmax: Discussed with JK offline, embedded webview will return YES for the start method, to avoid the completion block being called twice :)

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,33 +26,213 @@
...
+}
+
+
++ (BOOL)setCurrentWebSession:(id<MSIDWebviewInteracting>)newWebSession
+{
+    if (!s_currentWebSession)
+    {
+        @synchronized([MSIDWebviewAuthorization class])
```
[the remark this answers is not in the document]
unpluggedk: Yes you are right, will change

IdentityCore/src/webview/MSIDWebviewAuthorization.m
unpluggedk: I've included the session cancelling as discussed before

IdentityCore/src/cache/accessor/MSIDSharedTokenCache.h
```diff
@@ -0,0 +1,116 @@
...
+#import "MSIDBrokerResponse.h"
+#import "MSIDRefreshableToken.h"
+
+@class MSIDAccessToken;
+@class MSIDRefreshToken;
+@class MSIDLegacySingleResourceToken;
+@class MSIDBaseToken;
+@class MSIDOauth2Factory;
+
+@interface MSIDSharedTokenCache : NSObject
```
oldalton: this was removed

IdentityCore/src/configuration/MSIDConfiguration.h
```diff
@@ -24,13 +24,19 @@
 #import <Foundation/Foundation.h>
 #import "MSIDNetworkConfiguration.h"
 
+@class MSIDPkce;
+@class MSIDClientInfo;
```
oldalton: nit: why is this necessary?

IdentityCore/src/oauth2/aad_v2/MSIDAADV2Oauth2Factory.m
```diff
@@ -177,17 +186,96 @@ - (BOOL)fillAccount:(MSIDAccount *)account
...
+    {
+        return configuration.explicitStartURL;
+    }
+    
+    NSURLComponents *urlComponents = [NSURLComponents new];
+    urlComponents.scheme = @"https";
+    
+    // get this from cache: authorizationendpoint if possible
+    urlComponents.host = configuration.authority.host;
+    urlComponents.path = [configuration.authority.path stringByAppendingString:MSID_OAUTH2_V2_AUTHORIZE_SUFFIX];
```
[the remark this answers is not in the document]
unpluggedk: Hence there is a comment in line 259 :)
going to change accordingly

IdentityCore/src/oauth2/aad_v2/MSIDAADV2Oauth2Factory.m
```diff
@@ -177,17 +186,96 @@ - (BOOL)fillAccount:(MSIDAccount *)account
...
+
+    NSMutableDictionary <NSString *, NSString *> *parameters = [self authorizationParametersFromConfiguration:configuration];
+    
+    if (configuration.sliceParameters)
```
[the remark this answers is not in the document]
unpluggedk: Will see if after looking at V1OAuthFactory implementation on how much of this we could separate.

IdentityCore/src/util/NSString+MSIDExtensions.m
```diff
@@ -327,17 +329,36 @@ - (NSString *)msidUrlFormEncode
...
-    NSMutableString* toReturn = [[NSMutableString alloc] initWithCapacity:CC_SHA256_DIGEST_LENGTH*2];
-    for (int i = 0; i < sizeof(hash)/sizeof(hash[0]); ++i)
+    NSData *inputData = [self dataUsingEncoding:NSASCIIStringEncoding];
+    NSMutableData *outData = [NSMutableData dataWithLength:CC_SHA256_DIGEST_LENGTH];
+    
+    // input length shouldn't be this big
+    if (inputData.length > UINT32_MAX)
     {
-        [toReturn appendFormat:@"%02x", hash[i]];
+        @throw [NSException exceptionWithName: NSInvalidArgumentException
```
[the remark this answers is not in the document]
unpluggedk: 👍

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,166 @@
...
+    self = [super init];
+    if (self)
+    {
+        _startURL = url;
+        _context = context;
+        _requestState = requestState;
+        _stateVerifier = stateVerifier;
+        
+        _safariViewController = [[SFSafariViewController alloc] initWithURL:url entersReaderIfAvailable:NO];
+        _safariViewController.delegate = self;
```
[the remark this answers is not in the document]
unpluggedk: nope, it's weak

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,166 @@
...
+    
+    _completionHandler = completionHandler;
+
+    _telemetryRequestId = [_context telemetryRequestId];
+    
+    [[MSIDTelemetry sharedInstance] startEvent:_telemetryRequestId eventName:MSID_TELEMETRY_EVENT_UI_EVENT];
+    _telemetryEvent = [[MSIDTelemetryUIEvent alloc] initWithName:MSID_TELEMETRY_EVENT_UI_EVENT
+                                                         context:_context];
+    
+    [_telemetryEvent setIsCancelled:NO];
```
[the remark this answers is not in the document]
unpluggedk: ok 👍

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,166 @@
...
+    
+    dispatch_async(dispatch_get_main_queue(), ^{
```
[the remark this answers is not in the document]
unpluggedk: we'll make it consistently just run on main thread

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,166 @@
...
+- (void)cancel
+{
+    NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorSessionCanceled, @"Authorization session was cancelled programatically", nil, nil, nil, _context.correlationId, nil);
+    
+    [self completeSessionWithResponse:nil context:_context error:error];
+}
+
+- (BOOL)startWithCompletionHandler:(MSIDWebUICompletionHandler)completionHandler
+{
+    UIViewController *viewController = [UIApplication msidCurrentViewController];
```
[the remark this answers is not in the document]
unpluggedk: I think we did it to add in a webview in ADAL. 
For system webviews, it's presentViewController, and it'd make sense to just get the current root and present it.

IdentityCore/src/webview/systemWebview/MSIDSafariViewController.m
```diff
@@ -0,0 +1,166 @@
...
+    _completionHandler(response, authError);
+    return YES;
+}
+
+
+#pragma mark - SFSafariViewControllerDelegate
+- (void)safariViewControllerDidFinish:(SFSafariViewController *)controller
+{
+    // user cancel
+    NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorUserCancel, @"User cancelled the authorization session.", nil, nil, nil, _context.correlationId, nil);
```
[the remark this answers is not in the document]
unpluggedk: 👍

IdentityCore/tests/MSIDWebWPJResponseTests.m
```diff
@@ -0,0 +1,71 @@
...
+}
+
+
+- (void)testInit_whenGoodInput_shouldReturnResponsewithNoError
+{
+    NSError *error = nil;
+    MSIDWebWPJAuthResponse *response = [[MSIDWebWPJAuthResponse alloc] initWithScheme:@"msauth"
+                                                                           parameters:@{
+                                                                                        @"app_link":@"https://link",
+                                                                                        @"upn":@"user@sample.com"
```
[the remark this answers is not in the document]
unpluggedk: scheme is always msauth, 
however, what I missed out on is check wpj or broker as a host in the url.

IdentityCore/tests/util/cache/MSIDTestRequestParams.m
```diff
@@ -0,0 +1,76 @@
...
+// THE SOFTWARE.
+
+#import "MSIDTestRequestParams.h"
+#import "MSIDRequestParameters.h"
+#import "MSIDTestCacheIdentifiers.h"
+#import "NSOrderedSet+MSIDExtensions.h"
+
+@implementation MSIDTestRequestParams
+
++ (MSIDRequestParameters *)defaultParams
```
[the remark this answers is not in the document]
unpluggedk: it's from the prj merging 😂 
it's not in the project but the file remained.

IdentityCore/src/cache/accessor/MSIDSharedTokenCache.h
unpluggedk: it was removed from the project, but the file stayed. go merging 😂

IdentityCore/src/webview/systemWebview/MSIDSFAuthenticationSession.m
```diff
@@ -0,0 +1,119 @@
...
+    
+    if (@available(iOS 11.0, *))
+    {
+        _authSession = [[SFAuthenticationSession alloc] initWithURL:_startURL
+                                                  callbackURLScheme:_callbackURLScheme
+                                                  completionHandler:^(NSURL * _Nullable callbackURL, NSError * _Nullable error)
+                                            {
+                                                if (error)
+                                                {
+                                                    if (error.code == SFAuthenticationErrorCanceledLogin)
```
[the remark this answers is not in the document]
unpluggedk: Yes and added.

IdentityCore/src/webview/MSIDWebviewAuthorization.h
```diff
@@ -26,24 +26,56 @@
...
+                                     completionHandler:(MSIDWebviewAuthCompletionHandler)completionHandler;
+#endif
+
++ (BOOL)setCurrentWebSession:(MSIDWebviewSession *)session;
++ (void)cancelCurrentSession;
 
-+ (id<MSIDWebviewInteracting>)systemWebviewControllerWithConfiguration:(MSIDConfiguration *)configuration
-                                                               factory:(MSIDOAuth2Factory *)factory;
+// This is for system webview auth session on iOS 10 - Thus, a SafariViewController
++ (BOOL)handleURLResponseForSystemWebviewController:(NSURL *)url;
```
oldalton: nit: should this be iOS only?

IdentityCore/src/webview/MSIDWebviewAuthorization.h
```diff
@@ -26,24 +26,56 @@
...
+
+// This can be utilized for having a custom webview controller, and for testing.
++ (void)startSession:(MSIDWebviewSession *)session
+         verifyState:(BOOL)verifyState 
+             context:(id<MSIDRequestContext>)context
+   completionHandler:(MSIDWebviewAuthCompletionHandler)completionHandler;
+
+@property (class, readonly) MSIDWebviewSession *currentSession;
```
oldalton: nit: If you have a class property, can you make it writable and remove setCurrentWebSession?

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,34 +26,176 @@
...
+}
+
+
++ (BOOL)handleURLResponseForSystemWebviewController:(NSURL *)url;
+{
+#if TARGET_OS_IPHONE
+    @synchronized([MSIDWebviewAuthorization class])
+    {
+        if (s_currentSession &&
+            [(NSObject *)s_currentSession.webviewController isKindOfClass:MSIDSystemWebviewController.class])
```
oldalton: nit: do you need to cast to NSObject?

IdentityCore/src/webview/MSIDWebviewSession.h
```diff
@@ -0,0 +1,47 @@
...
+//------------------------------------------------------------------------------
+
+#import <Foundation/Foundation.h>
+
+@protocol MSIDWebviewInteracting;
+@class MSIDWebviewFactory;
+
+@interface MSIDWebviewSession : NSObject
+
+@property id<MSIDWebviewInteracting> webviewController;
```
oldalton: nit: consider using NSObject here instead

IdentityCore/src/webview/MSIDWebviewSession.m
```diff
@@ -0,0 +1,58 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+#import "MSIDWebviewSession.h"
+
+@implementation MSIDWebviewSession
+
+- (instancetype)initWithWebviewController:(id<MSIDWebviewInteracting>)webviewController
```
oldalton: this isn't public?

IdentityCore/src/webview/MSIDWebviewSession.m
```diff
@@ -0,0 +1,58 @@
...
+}
+
+
+- (instancetype)initWithWebviewController:(id<MSIDWebviewInteracting>)webviewController
+                                  factory:(MSIDWebviewFactory *)factory
+                             requestState:(NSString *)state
+                              verifyState:(BOOL)verifyState
+{
+    self = [super init];
+    if (self) {
```
oldalton: nit: { position

IdentityCore/src/webview/response/MSIDWebAADAuthResponse.h
```diff
@@ -27,8 +27,10 @@
 
 #import "MSIDWebOAuth2Response.h"
 
+@class MSIDWebOAuth2Response;
+
 @interface MSIDWebAADAuthResponse : MSIDWebOAuth2Response
 
-@property NSString *cloudHostName;
+@property(readonly) NSString *cloudHostName;
```
oldalton: nit: space

IdentityCore/src/webview/response/MSIDWebAADAuthResponse.h
```diff
@@ -27,8 +27,10 @@
 
 #import "MSIDWebOAuth2Response.h"
 
+@class MSIDWebOAuth2Response;
```
oldalton: nit: remove this

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,66 @@
...
+        // populate oauth error
+        _oauthError = oauthError;
+    }
+    
+    return self;
+}
+
++ (NSError *)oauthErrorFromParameters:(NSDictionary *)parameters
+{
+    NSUUID *correlationId = [parameters objectForKey:MSID_OAUTH2_CORRELATION_ID_RESPONSE] ?
```
oldalton: I think correlation ID is AAD specific thingy

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -26,8 +26,66 @@
...
+
+    if (serverOAuth2Error)
+    {
+        //login_required  ; has error_description
+        //access_denied ; has error_subcode
+        NSString *errorDescription = parameters[MSID_OAUTH2_ERROR_DESCRIPTION];
+        if (!errorDescription)
+        {
+            //
+            errorDescription = parameters[MSID_OAUTH2_ERROR_SUBCODE];
```
oldalton: is this subcode hack AAD specific?

IdentityCore/src/webview/response/MSIDWebWPJAuthResponse.m
```diff
@@ -29,4 +29,32 @@
...
 
+- (instancetype)initWithURL:(NSURL *)url
+                    context:(id<MSIDRequestContext>)context
+                      error:(NSError **)error
+{
+    NSString *scheme = url.scheme;
+    NSString *host = url.host;
+    
+    // Check for WPJ or broker response
+    if (!([scheme isEqualToString:@"msauth"] && [host isEqualToString:@"wpj"]))
```
oldalton: just verifying: where do we handle response, when scheme is msauth and host is either "code" or "microsoft.aad.brokerplugin"?

IdentityCore/src/webview/response/MSIDWebWPJAuthResponse.m
```diff
@@ -29,4 +29,32 @@
...
+    {
+        if (error){
```
oldalton: nit: { position

IdentityCore/src/webview/response/MSIDWebviewResponse.h
```diff
@@ -0,0 +1,41 @@
...
+//
+//------------------------------------------------------------------------------
+
+#import <Foundation/Foundation.h>
+
+@protocol MSIDRequestContext;
+
+@interface MSIDWebviewResponse : NSObject
+
+@property NSDictionary *parameters;
```
oldalton: should this be readonly, or we want it to be writable?

IdentityCore/src/webview/response/MSIDWebviewResponse.m
```diff
@@ -0,0 +1,64 @@
...
+@implementation MSIDWebviewResponse
+
+- (instancetype)initWithURL:(NSURL *)url
+                    context:(id<MSIDRequestContext>)context
+                      error:(NSError **)error
+{
+    if (!url)
+    {
+        if (error){
+            *error = MSIDCreateError(MSIDOAuthErrorDomain, MSIDErrorInvalidParameter, @"Trying to create a response with nil", nil, nil, nil, context.correlationId, nil);
```
oldalton: nit: with nil "url"

IdentityCore/src/webview/systemWebview/MSIDSFAuthenticationSession.m
```diff
@@ -0,0 +1,108 @@
...
+{
+    _telemetryRequestId = [_context telemetryRequestId];
+    [[MSIDTelemetry sharedInstance] startEvent:_telemetryRequestId eventName:MSID_TELEMETRY_EVENT_UI_EVENT];
+    _telemetryEvent = [[MSIDTelemetryUIEvent alloc] initWithName:MSID_TELEMETRY_EVENT_UI_EVENT
+                                                         context:_context];
+    
+    if (@available(iOS 11.0, *))
+    {
+        _authSession = [[SFAuthenticationSession alloc] initWithURL:_startURL
+                                                  callbackURLScheme:_callbackURLScheme
```
oldalton: do we pass full redirectUri in here?

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
```diff
@@ -26,19 +26,100 @@
...
+#import "MSIDOauth2Factory.h"
 
 @implementation MSIDSystemWebviewController
+{
+    id<MSIDRequestContext> _context;
+
+    MSIDSFAuthenticationSession *_authSession;
+    MSIDSafariViewController *_safariViewController;
+    
+    MSIDOauth2Factory *_factory;
```
oldalton: where's factory being set?

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
```diff
@@ -26,19 +26,100 @@
...
+    }
+    
+    NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractiveSessionStartFailure, @"Failed to create an auth session", nil, nil, nil, _context.correlationId, nil);
+    completionHandler(nil, error);
 }
 
+
 - (void)cancel
 {
+    if (@available(iOS 11.0, *))
```
oldalton: nit: both session and safariViewController conform to the same protocol, don't they?

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
```diff
@@ -26,19 +26,100 @@
...
+
+- (BOOL)handleURLResponseForSafariViewController:(NSURL *)url
+{
+    if (!url)
+    {
+        MSID_LOG_ERROR(_context, @"nil passed into the MSID Web handle response.");
+        return NO;
+    }
+    
+    if (@available(iOS 11.0, *)) { return NO; }
```
oldalton: nit: should this be yes as we don't have anything to handle?

IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
```diff
@@ -27,25 +27,39 @@
...
+// State verifier: Recommended verifier for state value of the response.
+//  Set to YES to stop if verifying state fails
+@property (readwrite) BOOL verifyState;
 
+// PKCE Support
 @property (readwrite) MSIDPkce *pkce;
-@property (readwrite) MSIDClientInfo *clientInfo;
+
+// User information
+@property (readwrite) NSString *utid;
```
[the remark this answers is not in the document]
unpluggedk: it is used in the param settings for v2

IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
```diff
@@ -27,25 +27,39 @@
...
+@property (readwrite) NSString *uid;
 
 // Priority start URL
 @property (readwrite) NSURL *explicitStartURL;
 
+- (instancetype)initWithAuthority:(NSURL *)authority
```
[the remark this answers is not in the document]
unpluggedk: changed according to offline discussion : MSIDWebviewConfig is different enough to be its own NSObject.

IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
```diff
@@ -27,25 +27,39 @@
...
+            authorizationEndpoint:(NSURL *)authorizationEndpoint
+                      redirectUri:(NSString *)redirectUri
+                         clientId:(NSString *)clientId
+                           target:(NSString *)target
+                    correlationId:(NSUUID *)correlationId;
+
+- (instancetype)initWithAuthority:(NSURL *)authority redirectUri:(NSString *)redirectUri clientId:(NSString *)clientId target:(NSString *)target correlationId:(NSUUID *)correlationId NS_UNAVAILABLE;
+- (instancetype)initWithAuthority:(NSURL *)authority redirectUri:(NSString *)redirectUri clientId:(NSString *)clientId target:(NSString *)target NS_UNAVAILABLE;
```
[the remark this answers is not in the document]
unpluggedk: Will be removed - no longer a subclass.

IdentityCore/src/webview/systemWebview/MSIDSFAuthenticationSession.m
unpluggedk: Scheme here should be enough

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
unpluggedk: nowhere, removed

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
unpluggedk: yeah, I could change this to remove this unnecessary check.

IdentityCore/src/webview/systemWebview/MSIDSystemWebviewController.m
unpluggedk: no, it should be no if we don't handle it.
think about it from this angle, 
it app delegate, it will do something like
return microsoft_url_handle || other_url_handle

IdentityCore/src/oauth2/MSIDOauth2Factory.h
```diff
@@ -59,9 +61,9 @@
 - (MSIDLegacySingleResourceToken *)legacyTokenFromResponse:(MSIDTokenResponse *)response configuration:(MSIDConfiguration *)configuration;
 - (MSIDAccount *)accountFromResponse:(MSIDTokenResponse *)response configuration:(MSIDConfiguration *)configuration;
 
-// Webviews
-- (id<MSIDWebviewInteracting>)embeddedWebviewControllerWithRequest:(MSIDConfiguration *)requestParams
-                                                     customWebview:(WKWebView *)webview;
-- (id<MSIDWebviewInteracting>)systemWebviewControllerWithRequest:(MSIDConfiguration *)requestParams;
+// Webview related
+- (MSIDWebviewFactory *)webviewFactory;
```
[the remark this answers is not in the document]
unpluggedk: sure, readonly property seems legit here.

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+#import <WebKit/WebKit.h>
+#import "MSIDSystemWebviewController.h"
+
+@implementation MSIDWebviewFactory
+
+#pragma mark - Webview creation
+
+- (MSIDWebviewSession *)embeddedWebviewSessionFromConfiguration:(MSIDWebviewConfiguration *)configuration verifyState:(BOOL)verifyState customWebview:(WKWebView *)webview context:(id<MSIDRequestContext>)context
+{
+    return nil;
```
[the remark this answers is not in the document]
unpluggedk: Yes

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+    {
+        [parameters addEntriesFromDictionary:configuration.sliceParameters];
+    }
+    
+    if (configuration.extraQueryParameters)
+    {
+        [parameters addEntriesFromDictionary:configuration.extraQueryParameters];
+    }
+    
+    parameters[MSID_OAUTH2_SCOPE] = MSID_OAUTH2_SCOPE_OPENID_VALUE;
```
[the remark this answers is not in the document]
unpluggedk: i will add to v2 factory as v1 doesn't support this.

IdentityCore/src/oauth2/aad_v2/MSIDAADV2WebviewFactory.m
```diff
@@ -0,0 +1,73 @@
...
+}
+
+
+- (NSMutableDictionary<NSString *,NSString *> *)authorizationParametersFromConfiguration:(MSIDWebviewConfiguration *)configuration requestState:(NSString *)state
+{
+    NSMutableDictionary<NSString *, NSString *> *parameters = [super authorizationParametersFromConfiguration:configuration
+                                                                                                 requestState:state];
+    
+    NSOrderedSet<NSString *> *allScopes = configuration.scopes;
+    parameters[MSID_OAUTH2_SCOPE] = [NSString stringWithFormat:@"%@ %@", MSID_OAUTH2_SCOPE_OPENID_VALUE,  [allScopes msidToString]];
```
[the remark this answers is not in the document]
unpluggedk: I'm fine with either, it's not a mutable object to begin with, I would expect performance/memory to have negligible difference. 
Aesthetic purposes though I think it might look better.

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+    parameters[MSID_OAUTH2_CLIENT_ID] = configuration.clientId;
+    parameters[MSID_OAUTH2_RESPONSE_TYPE] = MSID_OAUTH2_CODE;
+    parameters[MSID_OAUTH2_REDIRECT_URI] = configuration.redirectUri;
+    parameters[MSID_OAUTH2_CORRELATION_ID_REQUEST] = [configuration.correlationId UUIDString];
```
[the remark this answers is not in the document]
unpluggedk: Yes, will move it to base aad factory

IdentityCore/src/oauth2/aad_v2/MSIDAADV2WebviewFactory.m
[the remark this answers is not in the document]
unpluggedk: Actually, I think it makes more sense both functionally and aesthetically, to append to the allScopes (set) than a string manipulation.

IdentityCore/src/oauth2/aad_v2/MSIDAADV2WebviewFactory.m
```diff
@@ -0,0 +1,73 @@
...
+    return nil;
+}
+
+
+- (NSMutableDictionary<NSString *,NSString *> *)authorizationParametersFromConfiguration:(MSIDWebviewConfiguration *)configuration requestState:(NSString *)state
+{
+    NSMutableDictionary<NSString *, NSString *> *parameters = [super authorizationParametersFromConfiguration:configuration
+                                                                                                 requestState:state];
+    
+    NSOrderedSet<NSString *> *allScopes = configuration.scopes;
```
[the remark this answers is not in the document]
unpluggedk: I've changed it so we append to the set, not a string manipulation.

IdentityCore/src/oauth2/aad_v1/MSIDAADV1WebviewFactory.m
```diff
@@ -0,0 +1,43 @@
...
+{
+    return nil;
+}
+
+- (NSMutableDictionary<NSString *,NSString *> *)authorizationParametersFromConfiguration:(MSIDWebviewConfiguration *)configuration requestState:(NSString *)state
+{
+    NSMutableDictionary<NSString *, NSString *> *parameters = [super authorizationParametersFromConfiguration:configuration
+                                                                                                 requestState:state];
+    
+    // Do custom parameters for AAD V1
```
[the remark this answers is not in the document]
unpluggedk: This will be sorted with embedded webview pr

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+- (MSIDWebviewResponse *)responseWithURL:(NSURL *)url
+                            requestState:(NSString *)requestState
+                             verifyState:(BOOL)verifyState
+                                 context:(id<MSIDRequestContext>)context
+                                   error:(NSError **)error
+{
+    NSError *stateVerifierError = nil;
+    if (![self verifyRequestState:requestState responseURL:url error:&stateVerifierError] && verifyState)
+    {
+        if (error)
```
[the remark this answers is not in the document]
unpluggedk: sure

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+    // Check for auth response
+    // Try both the URL and the fragment parameters:
+    NSDictionary *parameters = [url msidFragmentParameters];
+    if (parameters.count == 0)
+    {
+        parameters = [url msidQueryParameters];
+    }
+
+    NSString *stateReceived = parameters[MSID_OAUTH2_STATE];
+    BOOL result = [requestState isEqualToString:stateReceived.msidBase64UrlDecode];
```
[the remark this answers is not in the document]
unpluggedk: The value we generate is still in the discussion, so for us to not break anything regardless of what we decide to set as a request state, I've put in encoding for the request state, and thus decoding for the received state.

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -0,0 +1,183 @@
...
+                                     nil, nil, nil, nil, nil);
+        }
+    }
+    
+    return result;
+}
+
+- (NSString *)generateStateValue
+{
+    return [[NSUUID UUID] UUIDString];
```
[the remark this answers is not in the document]
unpluggedk: Almost. Writing up a formal doc to finalize it.

IdentityCore/src/oauth2/aad_v2/MSIDAADV2WebviewFactory.m
```diff
@@ -0,0 +1,73 @@
...
+    parameters[@"haschrome"] = @"1";
+    
+    return parameters;
+}
+
+- (MSIDWebviewResponse *)responseWithURL:(NSURL *)url
+                                 context:(id<MSIDRequestContext>)context
+                                   error:(NSError **)error
+{
+    MSIDWebWPJAuthResponse *wpjResponse = [[MSIDWebWPJAuthResponse alloc] initWithURL:url context:context error:nil];
```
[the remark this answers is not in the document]
unpluggedk: good point. why don't we!!?

(now we do.)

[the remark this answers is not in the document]
unpluggedk: But, after writing this, I think it would be good to actually use an error pointer here and at least log why it failed to create wpj response.

IdentityCore/src/oauth2/aad_v2/MSIDAADV2WebviewFactory.m
```diff
@@ -0,0 +1,73 @@
...
+
+    if (wpjResponse) return wpjResponse;
+    
+    NSError *responseCreationError = nil;
+    MSIDWebAADAuthResponse *response = [[MSIDWebAADAuthResponse alloc] initWithURL:url
+                                                                           context:context
+                                                                             error:&responseCreationError];
+    if (responseCreationError) {
+        if (error)  *error = responseCreationError;
```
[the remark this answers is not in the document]
unpluggedk: in this scenario, I think yes, for WPJ creation, no because I don't want to have an error after AAD auth response creation succeeds

IdentityCore/src/webview/MSIDWebviewAuthorization.h
unpluggedk: sure

IdentityCore/src/webview/MSIDWebviewAuthorization.h
unpluggedk: setting session was meant for testing only, I will actually remove it from the header file.

IdentityCore/src/webview/MSIDWebviewAuthorization.m
unpluggedk: yeah because .webviewController is an id<..>.

IdentityCore/src/webview/MSIDWebviewSession.h
unpluggedk: let us do that.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: moved accordingly

IdentityCore/src/webview/response/MSIDWebviewResponse.h
unpluggedk: I think it should be read-only. let me change that.

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
unpluggedk: It was, and let me remove this, 
we can pass "haschrome"=1 to disable back button altogether. Thus, no need to handle with hack

IdentityCore/src/webview/response/MSIDWebWPJAuthResponse.m
unpluggedk: Implicitly, it would fail to create WPJ here in this class.
After that, MSIDWebAADAuthResponse would be created to handle it.

For reference I've created #145 , due to importance of broker scenarios.

IdentityCore/src/util/NSString+MSIDExtensions.m
```diff
@@ -327,17 +329,35 @@ - (NSString *)msidUrlFormEncode
...
+//    const char* inputStr = [self UTF8String];
+//    unsigned char hash[CC_SHA256_DIGEST_LENGTH];
+//    CC_SHA256(inputStr, (int)strlen(inputStr), hash);
+//    NSMutableString* toReturn = [[NSMutableString alloc] initWithCapacity:CC_SHA256_DIGEST_LENGTH*2];
+//    for (int i = 0; i < sizeof(hash)/sizeof(hash[0]); ++i)
+//    {
+//        [toReturn appendFormat:@"%02x", hash[i]];
+//    }
+//    return toReturn;
+    return [NSString msidBase64UrlEncodeData:[self msalSHA256Data]];
```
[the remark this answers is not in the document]
unpluggedk: Let's decide on what the default should be for this. #147

IdentityCore/src/webview/MSIDWebviewSession.m
unpluggedk: it is now :)

IdentityCore/src/configuration/MSIDNetworkConfiguration.m
```diff
@@ -23,41 +23,28 @@
...
++ (void)setTimeout:(NSTimeInterval)timeout
 {
-    self = [super init];
-    if (self)
-    {
-        _timeout = s_defaultTimeout;
-        _retryCount = s_defaultRetryCount;
-    }
-    return self;
+    s_timeout = timeout;
```
antrix1989: Maybe it is better to have default _timeout_ & _retryCount_ setter & getters and assign default values to them in initializer? In such case you will not need to override setters & getters.

IdentityCore/src/configuration/MSIDNetworkConfiguration.h
```diff
@@ -24,11 +24,9 @@
 #import <Foundation/Foundation.h>
 
 
-@interface MSIDNetworkConfiguration : NSObject <NSCopying>
+@interface MSIDNetworkConfiguration : NSObject
```
antrix1989: Should MSIDNetworkConfiguration be singleton? If no, how are we going to access it in the code?

unpluggedk: Both it's properties are class properties with setters and getters.
So, you'd just use MSIDNetworkConfiguration.timeout and MSIDNetworkConfiguration.retryCount

IdentityCore/src/configuration/MSIDNetworkConfiguration.m
unpluggedk: these are class properties, which means they don't have default synthesizer.
thus, we'd need to write a setter/getter anyways. so, with that, I don't see a need for an initializer.

IdentityCore/src/configuration/MSIDNetworkConfiguration.h
unpluggedk: https://developer.apple.com/videos/play/wwdc2016/405/  at 5:07 for reference.

IdentityCore/src/webview/MSIDWebviewAuthorization.h
unpluggedk: Setting it needs to return a BOOL as it can only be set when there is no session running, thus a need for a separate setter.

IdentityCore/src/util/NSString+MSIDExtensions.m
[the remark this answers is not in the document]
unpluggedk: Being addressed in #149

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
```diff
@@ -0,0 +1,88 @@
...
+         @{
+           MSID_OAUTH2_CORRELATION_ID_REQUEST : @"true",
+           MSID_OAUTH2_CORRELATION_ID_REQUEST_VALUE : [configuration.correlationId UUIDString]
+           }];
+    }
+    
+    if (configuration.sliceParameters)
+    {
+        [parameters addEntriesFromDictionary:configuration.sliceParameters];
+    }
```
jasoncoolmax: We already add slice parameters at line 38 :)

IdentityCore/src/webview/MSIDWebviewAuthorization.m
```diff
@@ -26,34 +26,185 @@
...
+    
+    [self startSession:session verifyState:verifyState context:context completionHandler:completionHandler];
+}
+#endif
+
++ (void)startSession:(MSIDWebviewSession *)session
+         verifyState:(BOOL)verifyState 
+             context:(id<MSIDRequestContext>)context
+   completionHandler:(MSIDWebviewAuthCompletionHandler)completionHandler
+{
```
jasoncoolmax: parameter `verifyState` is no longer used in this function anymore :)

IdentityCore/IdentityCore.xcodeproj/project.pbxproj
```diff
@@ -2823,7 +3042,7 @@
 				GCC_WARN_UNUSED_FUNCTION = YES;
 				GCC_WARN_UNUSED_VARIABLE = YES;
 				INFOPLIST_FILE = MSIDTestsHostApp/Info.plist;
-				IPHONEOS_DEPLOYMENT_TARGET = 9.0;
+				IPHONEOS_DEPLOYMENT_TARGET = 10.3;
```
oldalton: Why did we change deployment target?

IdentityCore/src/MSIDError.h
oldalton: nit: can we put errors in increasing order? (MSIDErrorInvalidState is in the middle)

IdentityCore/src/MSIDError.h
```diff
@@ -68,11 +68,53 @@ typedef NS_ENUM(NSInteger, MSIDErrorCode)
...
+    
     MSIDErrorUnsupportedFunctionality = -51018,
+
```
oldalton: both MSIDErrorInvalidParameter = -51018, and MSIDErrorUnsupportedFunctionality = -51018 are -51018?

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
```diff
@@ -0,0 +1,88 @@
...
+    NSMutableDictionary<NSString *, NSString *> *parameters = [super authorizationParametersFromConfiguration:configuration
+                                                                                                 requestState:state];
+    
+    if (configuration.sliceParameters)
+    {
+        [parameters addEntriesFromDictionary:configuration.sliceParameters];
+    }
+
+    NSMutableOrderedSet<NSString *> *allScopes = parameters[MSID_OAUTH2_SCOPE].scopeSet.mutableCopy;
+    [allScopes addObject:MSID_OAUTH2_SCOPE_OPENID_VALUE];
```
jasoncoolmax: `allScopes` will be `nil` if no scopes are set? :)

IdentityCore/src/configuration/MSIDConfiguration.m
```diff
@@ -33,27 +34,15 @@ - (instancetype)copyWithZone:(NSZone*)zone
     configuration.redirectUri = [_redirectUri copyWithZone:zone];
     configuration.target = [_target copyWithZone:zone];
     configuration.clientId = [_clientId copyWithZone:zone];
-    configuration.correlationId = [_correlationId copyWithZone:zone];
-    configuration.loginHint = [_loginHint copyWithZone:zone];
-    configuration.sliceParameters = [_sliceParameters copyWithZone:zone];
```
oldalton: I think we need slice parameters for all requests, both interactive and silent

IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
```diff
@@ -26,26 +26,46 @@
...
 @property (readwrite) NSDictionary<NSString *, NSString *> *extraQueryParameters;
+@property (readwrite) NSDictionary<NSString *, NSString *> *sliceParameters;
 @property (readwrite) NSString *promptBehavior;
 @property (readwrite) NSString *claims;
 
-// Is this only for V2?
-@property (readwrite) NSString *requestState;
+// State verifier: Recommended verifier for state value of the response.
+//  Set to YES to stop if verifying state fails
+@property (readwrite) BOOL verifyState;
```
oldalton: nit: I'd prefer verify state to be set on initialize, so nobody can change it in the middle of authentication.

IdentityCore/src/oauth2/MSIDOauth2Factory.h
```diff
@@ -59,9 +64,8 @@
 - (MSIDLegacySingleResourceToken *)legacyTokenFromResponse:(MSIDTokenResponse *)response configuration:(MSIDConfiguration *)configuration;
 - (MSIDAccount *)accountFromResponse:(MSIDTokenResponse *)response configuration:(MSIDConfiguration *)configuration;
 
-// Webviews
-- (id<MSIDWebviewInteracting>)embeddedWebviewControllerWithRequest:(MSIDConfiguration *)requestParams
-                                                     customWebview:(WKWebView *)webview;
-- (id<MSIDWebviewInteracting>)systemWebviewControllerWithRequest:(MSIDConfiguration *)requestParams;
+// Webview related
+@property(readonly) MSIDWebviewFactory *webviewFactory;
```
oldalton: nit: space

IdentityCore/src/oauth2/MSIDWebviewFactory.m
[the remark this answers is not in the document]
oldalton: Oki :)

IdentityCore/src/oauth2/aad_base/MSIDAADOauth2Factory.m
```diff
@@ -231,4 +232,16 @@ - (BOOL)fillBaseToken:(MSIDBaseToken *)baseToken
     return YES;
 }
 
+
+#pragma mark - Webview
+- (MSIDWebviewFactory *)webviewFactory
+{
+    if (!_webviewFactory)
+    {
+        _webviewFactory = [[MSIDAADWebviewFactory alloc] init];
```
oldalton: nit: do we need to do it in each subclass?

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
oldalton: Good point, we should address it :)

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
```diff
@@ -0,0 +1,88 @@
...
+    
+    parameters[MSID_OAUTH2_SCOPE] = allScopes.msidToString;
+    parameters[MSID_OAUTH2_PROMPT] = configuration.promptBehavior;
+    
+    if (configuration.correlationId)
+    {
+        [parameters addEntriesFromDictionary:
+         @{
+           MSID_OAUTH2_CORRELATION_ID_REQUEST : @"true",
```
oldalton: I'm just curious why is this necessary?

IdentityCore/src/util/NSDictionary+MSIDExtensions.m
```diff
@@ -94,4 +94,22 @@ - (NSDictionary *)dictionaryByRemovingFields:(NSArray *)fieldsToRemove
     return mutableDict;
 }
 
+- (NSArray<NSURLQueryItem *> *)urlQueryItemsArray;
```
oldalton: just curious, where is this being used?
and nit: there's a ";" in the end

IdentityCore/src/util/NSString+MSIDExtensions.m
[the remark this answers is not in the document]
oldalton: ok :)

IdentityCore/tests/MSIDWebOAuth2ResponseTests.m
```diff
@@ -0,0 +1,92 @@
...
+    NSError *error = nil;
+    NSString *errorString = @"invalid_grant";
+    NSString *errorDescription = @"error description";
+    NSString *subError = @"suberror";
+
+    NSURLComponents *urlComponents = [NSURLComponents componentsWithString:@"https://contoso.com"];
+    urlComponents.queryItems = @{
+                                 MSID_OAUTH2_ERROR : errorString,
+                                 MSID_OAUTH2_ERROR_DESCRIPTION : errorDescription,
+                                 MSID_OAUTH2_SUB_ERROR : subError,
```
oldalton: do we also have a test for the "subside" scenario?

IdentityCore/tests/MSIDAADV2Oauth2FactoryTests.m
```diff
@@ -417,5 +421,112 @@ - (void)testAccountFromTokenResponse_whenAADV2TokenResponse_shouldInitAccountAnd
     XCTAssertEqualObjects(account.authority.absoluteString, @"https://login.microsoftonline.com/contoso.com");
 }
 
+#pragma mark - Webview (StartURL)
+//- (void)testStartURL_whenValidParams_shouldContainQPs
```
oldalton: if this code is not necessary, let's remove it :)

IdentityCore/src/webview/MSIDWebviewAuthorization.m
unpluggedk: Let's see if there is a need for anyone to override this value from default.

IdentityCore/IdentityCore.xcodeproj/project.pbxproj
unpluggedk: 😕

IdentityCore/src/MSIDError.h
unpluggedk: it's magic. now you see it, now you don't.

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
unpluggedk: done

IdentityCore/src/configuration/MSIDConfiguration.m
unpluggedk: I think it is to be determined whether it will use MSIDConfiguration class or not.

IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
unpluggedk: sounds logical. will change accordingly.

IdentityCore/src/oauth2/aad_base/MSIDAADOauth2Factory.m
unpluggedk: they create different type of webview factory

IdentityCore/tests/MSIDWebOAuth2ResponseTests.m
unpluggedk: I don't think so. The only place where it made sense was when a developer clicks on a back button on the webview itself.
By passing haschrome=1, there no longer is the back button and thus, we are fine without.

IdentityCore/src/util/NSDictionary+MSIDExtensions.m
unpluggedk: we use this constructing startURL.
```
urlComponents.queryItems = [parameters urlQueryItemsArray];
```

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
unpluggedk: It was in x-plat libraries as well as our ADAL.

From MSDN, https://msdn.microsoft.com/en-us/library/mt739779.aspx

2.2.1.2 return-client-request-id
The return-client-request-id HTTP header is optional. This header is sent in the request and is used by the key provisioning server to determine whether to return the client-specified client-request-id in the server response. If present, the value of the return-client-request-id header MUST be "true".

The format of the return-client-request-id header, in ABNF, is as follows.

 return-client-request-id = "true"

IdentityCore/src/MSIDError.h
unpluggedk: Let's address errors in a separate PR

PR #228, opened 2018-09-14
Fix default handling of challenge twice in ClientTLS flow.

unpluggedk:


IdentityCore/src/webview/embeddedWebview/challangeHandlers/MSIDClientTLSHandler.m
```diff
@@ -56,7 +56,6 @@ + (BOOL)handleChallenge:(NSURLAuthenticationChallenge *)challenge
         return [self handleWPJChallenge:challenge context:context completionHandler:completionHandler];
     }
 #if TARGET_OS_IPHONE
-    completionHandler(NSURLSessionAuthChallengePerformDefaultHandling, nil);
     return NO;
```
jasoncoolmax: I think the current design is that, completionHandler is passed in, it should be called no matter we are returning YES or NO. Otherwise the challenge is not handled outside :)

jasoncoolmax: Oh, after second check, found that Olga has changed the logic here (https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/commit/526ea869a292254d22450051f80d9f4da3f51d48#diff-76788ab45efaf6db1829e7a545e1a199). 

So my above comment is longer true. Then the fix is needed :)

jasoncoolmax: I suggest to check all the challenge handlers for the case of returning NO, I remember I made sure completion handler USB called for all NO cases :)

unpluggedk: Yeah, did added some for PKeyAuth

PR #244, opened 2018-09-26
Fix Certificate Choose Crash

jasoncoolmax:
The CBA chooser crashes because it shouldn't nil out the caller in the callback.
I.e., it calls the following with the caller being `_panel` and callback being `sheetDidEnd:returnCode:contextInfo:`:
```
[_panel beginSheetForWindow:_window
                   modalDelegate:self
                  didEndSelector:@selector(sheetDidEnd:returnCode:contextInfo:)	                  
                     contextInfo:NULL
                      identities:identities
                         message:message];
```
(line 57).
 
However, `_panel`is nil out right in the callback `sheetDidEnd:returnCode:contextInfo:` (at line 90).

XCode Instruments showed that, `@selector(_okClicked:)` was sent to a deallocated `_panel`, which means even after/during the callback, system is still doing something to `_panel`. So we shouldn't nil it out.
(Tested on macOS 0.11, 10.12, 10.13 and 10.14)

Besides, we forgot to update notification using the MSIDNotifications. Now fixed.

### DIFF
--- a/IdentityCore/src/webview/embeddedWebview/ui/mac/MSIDCertificateChooser.m
+++ b/IdentityCore/src/webview/embeddedWebview/ui/mac/MSIDCertificateChooser.m
@@ -87,7 +87,6 @@
     }
     
     SecIdentityRef identity = _panel.identity;
-    _panel = nil;
     _completionHandler(identity);
     _completionHandler = nil;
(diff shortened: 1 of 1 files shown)


unpluggedk:
into /dev please

PR #245, opened 2018-09-27
minor fix for hexString length

rohitnarula7176:
*Minor fix in api which generates hex string from NSData


unpluggedk:
into /dev please

PR #268, opened 2018-10-25
Intune cache

antrix1989:


IdentityCore/src/MSIDJsonSerializer.h
```diff
@@ -0,0 +1,55 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+#import <Foundation/Foundation.h>
+
+@protocol MSIDJsonSerializable;
+@protocol MSIDRequestContext;
+
+@protocol MSIDJsonSerializing <NSObject>
```
unpluggedk: This is the only class that uses this protocol, what is the rational behind using protocols instead of simply declaring as an instance method directly?

IdentityCore/src/MSIDCache.h
```diff
@@ -35,6 +37,8 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (void)removeAllObjects;
 
+- (NSDictionary *)toDictionary;
```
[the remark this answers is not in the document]
antrix1989: I prefer to keep it because it indicates that it is not just a reference to the previously provided dictionary, but a new dictionary.

IdentityCore/src/MSIDJsonSerializer.h
antrix1989: done

PR #298, opened 2018-11-19
V2 Broker Protocol

jasoncoolmax:


IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+        NSString *userInfokey = _userInfoKeyMapping[metadataKey];
+        if (userInfokey)
+        {
+            [userInfo setValue:errorResponse.errorMetadata[metadataKey] forKey:userInfokey];
+        }
+    }
+
+    userInfo[MSIDUserDisplayableIdkey] = errorResponse.userId ? errorResponse.userId : userId;
+
+    NSDictionary *httpHeaders = [NSDictionary msidDictionaryFromWWWFormURLEncodedString:errorResponse.httpHeaders];
```
oldalton: Wouldn't http headers be part of error metadata?

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+    if (httpHeaders)
+        userInfo[MSIDHTTPHeadersKey] = httpHeaders;
+    
+    if (errorResponse.brokerAppVer)
```
oldalton: nit: no need to check for presence of brokerAppVer before setting it

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerTokenRequest.m
```diff
@@ -22,20 +22,34 @@
...
 - (NSDictionary *)protocolPayloadContentsWithError:(NSError **)error
 {
-    // TODO: MSAL pieces
-    return @{};
+    NSString *homeAccountId = self.requestParameters.accountIdentifier.homeAccountId;
+    NSString *username = self.requestParameters.accountIdentifier.legacyAccountId;
+    
+    NSDictionary *contents =
+    @{
+      @"scope": self.requestParameters.target ?: @"",
```
oldalton: Question for future extensibility: Do we want broker be responsible for appending OIDC scopes, or we want MSAL to drive OIDC scopes requested?

IdentityCore/tests/integration/MSIDDefaultAccessorSSOIntegrationTests.m
```diff
@@ -377,39 +377,6 @@ - (void)testSaveTokensWithFactory_whenMultiResourceFOCIResponse_savesTokensOnlyT
     XCTAssertEqual([allAccounts count], 0);
 }
 
-
-- (void)testSaveTokensWithFactoryAndBrokerResponse_shouldFailWithUnsupportedError
```
oldalton: Should there be a test instead of this one?

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
```diff
@@ -0,0 +1,488 @@
...
+    XCTAssertEqualObjects(result.accessToken.clientId, @"my_client_id");
+    XCTAssertEqualObjects(result.accessToken.accountIdentifier.homeAccountId, @"1.1234-5678-90abcdefg");
+    XCTAssertEqualObjects(result.accessToken.accountIdentifier.legacyAccountId, @"user@contoso.com");
+    XCTAssertEqualObjects(result.accessToken.authority.url.absoluteString, @"https://login.microsoftonline.com/contoso.com-guid");
+    XCTAssertTrue([expiresOn timeIntervalSinceDate:result.accessToken.expiresOn] < 1);
+    XCTAssertTrue([extExpiresOn timeIntervalSinceDate:result.accessToken.extendedExpireTime] < 1);
+    
+    XCTAssertEqualObjects(result.rawIdToken, idTokenString);
+    //Question:
+    //Should authority of MSIDTokenResult also be @"https://login.microsoftonline.com/contoso.com-guid"?
```
oldalton: Authority should be based on broker response.

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
```diff
@@ -0,0 +1,488 @@
...
+    XCTAssertEqualObjects(result.authority.url.absoluteString, @"https://login.microsoftonline.com/common");
+    XCTAssertEqualObjects(result.correlationId.UUIDString, correlationId);
+    //Question:
+    //We never set the following property of MSIResult, do we want to set it or remove it
```
oldalton: We should set in the silent token request for extended expires cases. It's a bug if we don't set it there, let me check. Can new broker return extended expiry tokens? If yes, we should set it for MSAL broker responses too.

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
```diff
@@ -0,0 +1,488 @@
...
+    
+    NSString *correlationId = [[NSUUID UUID] UUIDString];
+    
+    NSDictionary *errorMetadata =
+    @{
+      @"http_response_code" : @200,
+      @"error_description" : @"Error occured",
+      @"oauth_error" : @"invalid_grant",
+      @"oauth_sub_error" : @"consent_required",
+      @"declined_scopes" : @"decliendScope1 decliendScope2",
```
oldalton: I think declined scopes handling should happen on MSAL side, otherwise MSAL will never cache those tokens and next request will result in interactive auth. Can you make sure the protocol and code accounts for those cases?

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
```diff
@@ -0,0 +1,488 @@
...
+    
+    NSDictionary *intuneMAMTokenDictionary =
+    @{
+      @"authority" : @"https://login.microsoftonline.com/common",
+      @"scope" : scopes,
+      @"client_id" : @"my_client_id",
+      @"id_token" : mamIdTokenString,
+      @"client_info" : mamRawClientInfo,
+      @"access_token" : @"intune-mam-accesstoken",
+      @"token_type" : @"Bearer",
```
oldalton: Do we return token_type from broker?

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
```diff
@@ -0,0 +1,488 @@
...
+    
+    NSString *scopes = @"myscope1 myscope2";
+    
+    NSDictionary *intuneMAMTokenDictionary =
+    @{
+      @"authority" : @"https://login.microsoftonline.com/common",
+      @"scope" : scopes,
+      @"client_id" : @"my_client_id",
+      @"id_token" : mamIdTokenString,
+      @"client_info" : mamRawClientInfo,
```
oldalton: Should we also return homeAccountId, as it might be not based on client_info in some cases in the future, and client_info is also optional in schema.

IdentityCore/tests/util/MSIDKeychainTokenCache+MSIDTestsUtil.h
```diff
@@ -22,9 +22,19 @@
...
 #import "MSIDKeychainTokenCache.h"
+#import "MSIDCredentialType.h"
+
+@protocol MSIDCacheAccessor;
 
 @interface MSIDKeychainTokenCache (MSIDTestUtil)
 
 + (void)reset;
 
++ (NSArray *)getAllLegacyAccessTokens:(id<MSIDCacheAccessor>)cacheAccessor;
```
oldalton: 👍

IdentityCore/src/oauth2/aad_v2/MSIDAADV2BrokerResponse.h
```diff
@@ -27,13 +27,14 @@ NS_ASSUME_NONNULL_BEGIN
 
 @interface MSIDAADV2BrokerResponse : MSIDBrokerResponse
 
-@property (readonly) NSString *scope;
-@property (readonly) NSDictionary *errorUserInfo;
+//@property (readonly) NSString *scope;
```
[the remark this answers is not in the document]
jasoncoolmax: My bad, should have removed those two slashes :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+    {
+        errorDescription = @"Broker did not provide any details";
+    }
+    
+    NSString *oauthErrorCode = errorResponse.oauthErrorCode;
+    NSString *subError = errorResponse.subError;
+    NSUUID *correlationId = [[NSUUID alloc] initWithUUIDString:errorResponse.correlationId];
+    
+    NSMutableDictionary *userInfo = [NSMutableDictionary new];
+    for (NSString * metadataKey in errorResponse.errorMetadata.allKeys)
```
unpluggedk: nit: *

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerTokenRequest.m
```diff
@@ -22,20 +22,34 @@
...
+      @"home_account_id": homeAccountId ?: @"",
+      @"username": username ?: @"",
```
unpluggedk: any reasons on why we send it for empty string? I think we should not send it for them

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+    NSString *userDisplayableId = nil;
+    if (decryptedResponse[@"additional_tokens"])
+    {
+        MSIDTokenResult *tokenResult = nil;
+        NSError *additionalTokensError = nil;
+        
+        NSDictionary *additionalTokensDict = [decryptedResponse[@"additional_tokens"] msidJson];
+        if (additionalTokensDict)
+        {
+            MSIDAADV2BrokerResponse *brokerResponse = [[MSIDAADV2BrokerResponse alloc] initWithDictionary:additionalTokensDict error:&additionalTokensError];
```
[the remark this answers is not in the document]
jasoncoolmax: fixed :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+    }
+    
     return nil;
+#endif
 }
 
+- (NSError *)resultFromBrokerErrorResponse:(MSIDAADV2BrokerResponse *)errorResponse
+                         userDisplayableId:(NSString *)userId
+{
+    NSString *errorDomain = errorResponse.errorDomain ?: MSIDErrorDomain;
```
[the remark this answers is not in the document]
jasoncoolmax: You are right, i think we could remove the fallback as error_domain is mandatory for error cases :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -22,25 +22,184 @@
...
+        }
+        else
+        {
+            additionalTokensError = MSIDCreateError(MSIDErrorDomain, MSIDErrorBrokerCorruptedResponse, @"Unable to parse additional tokens.", nil, nil, nil, nil, nil);
+        }
+        
+        if (!tokenResult)
+        {
+            MSID_LOG_CORR_WARN(correlationID, @"Unable to save additional tokens with error %ld, %@", (long)additionalTokensError.code, additionalTokensError.domain);
+            MSID_LOG_WARN_CORR_PII(correlationID, @"Unable to save additional token with error %@", additionalTokensError);
```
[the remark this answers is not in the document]
oldalton: 1. We shouldn't fail the flow if we fail to save intune tokens, as we can still continue
2. This handles both error and success cases. We should be able to handle both.

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
jasoncoolmax: fixed :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
jasoncoolmax: The for-loop at line 181-187 can only handle string-type metadata, which is the case for most properties. Http headers are dictionary type, so need some special handling after the for-loop.

Added some comments to clarify the above :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
jasoncoolmax: 👍 fixed it :)

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
oldalton: Ok, but shouldn't it come from errorResponse.errorMetadata?

jasoncoolmax: Yep, the getter retrieves the value from `errorResponse.errorMetadata` :)

oldalton: 👍

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
jasoncoolmax: 👍

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
jasoncoolmax: Oh I See.
Nope, both new broker and old broker won't return extended expiry tokens, as all broker call is interactive :)

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
oldalton: Yes, makes sense :)

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
jasoncoolmax: Not by old broker. Does server returns anything other than Bearer tokens?

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerTokenRequest.m
jasoncoolmax: Talked to Olga offline. Client may want additional OIDC scopes, so we let client send the OIDC scopes. In broker, we will also check if the scope contains minimal OIDC scopes, add them if not.

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
jasoncoolmax: Made changes in the protocol doc to have "home_account_id" being returned from broker. 
Do we want to change the current common core logic to fetch home_account_id field from server; if unavailable, then fallback to client info?

IdentityCore/tests/integration/MSIDDefaultAccessorSSOIntegrationTests.m
jasoncoolmax: added :)

IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
jasoncoolmax: Discussed with Olga again, we decided to send declined_scopes as well as granted_scopes. Besides of that, we also send the tokens in the` additional_tokens` field.

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerTokenRequest.m
jasoncoolmax: fixed. Now only send if value is not empty :)

IdentityCore/src/MSIDError.h
```diff
@@ -28,7 +28,10 @@ extern NSString *MSIDCorrelationIdKey;
 extern NSString *MSIDHTTPHeadersKey;
 extern NSString *MSIDHTTPResponseCodeKey;
 extern NSString *MSIDUserDisplayableIdkey;
+extern NSString *MSIDHomeAccountIdkey;
 extern NSString *MSIDBrokerVersionKey;
+extern NSString *MSIDGrantedScopesKey;
+extern NSString *MSIDDeclinedScopesKey;
```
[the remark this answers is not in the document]
jasoncoolmax: :+1: fixed

IdentityCore/src/requests/sdk/adal/MSIDLegacyBrokerTokenRequest.m
```diff
@@ -53,15 +54,13 @@ - (NSDictionary *)protocolPayloadContentsWithError:(NSError **)error
...
-      @"resource": self.requestParameters.target ?: @"",
-      @"force": self.requestParameters.uiBehaviorType == MSIDUIBehaviorForceType ? @"YES" : @"NO",
-      @"username": username ?: @"",
-      @"username_type": usernameType,
-      @"max_protocol_ver": @"2"
-    };
+    NSMutableDictionary *contents = [NSMutableDictionary new];
+    [contents msidSetNonEmptyString:skipCacheValue forKey:@"skip_cache"];
+    [contents msidSetNonEmptyString:self.requestParameters.target forKey:@"resource"];
+    [contents msidSetNonEmptyString:self.requestParameters.uiBehaviorType == MSIDUIBehaviorForceType ? @"YES" : @"NO" forKey:@"force"];
```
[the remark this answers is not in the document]
jasoncoolmax: fixed :)

IdentityCore/src/requests/sdk/adal/MSIDLegacyBrokerTokenRequest.m
```diff
@@ -53,15 +54,13 @@ - (NSDictionary *)protocolPayloadContentsWithError:(NSError **)error
...
+    [contents msidSetNonEmptyString:username forKey:@"username"];
+    [contents msidSetNonEmptyString:usernameType forKey:@"username_type"];
+    [contents msidSetNonEmptyString:@"2" forKey:@"max_protocol_ver"];
```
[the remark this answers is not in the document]
jasoncoolmax: fixed

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerTokenRequest.m
```diff
@@ -22,20 +22,34 @@
...
 - (NSDictionary *)protocolPayloadContentsWithError:(NSError **)error
 {
-    // TODO: MSAL pieces
-    return @{};
+    NSString *homeAccountId = self.requestParameters.accountIdentifier.homeAccountId;
+    NSString *username = self.requestParameters.accountIdentifier.legacyAccountId;
+    
+    // if value is nil, it won't appear in the dictionary
+    NSMutableDictionary *contents = [NSMutableDictionary new];
+    [contents msidSetNonEmptyString:self.requestParameters.allTokenRequestScopes forKey:@"scope"];
```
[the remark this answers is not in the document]
jasoncoolmax: then probably we will need one more field to send `oidc_scopes`, what do you think?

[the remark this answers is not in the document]
oldalton: Makes sense. Would be a bit confusing that we have scope and oidc_scopes though. Maybe we can call "scope" as something else then (e.g. resource_scopes). Or call oidc_scopes something like extra_oidc_scopes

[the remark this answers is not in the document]
jasoncoolmax: Shall we avoid using `resource` to avoid confusion between V1 and V2? :) 
How about `request_scopes` and `extea_oidc_scopes`?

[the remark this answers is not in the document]
oldalton: request_scopes  and extra_oidc_scopes sound good to me :)

IdentityCore/src/util/NSMutableDictionary+MSIDExtensions.h
```diff
@@ -26,5 +26,6 @@
 @interface NSMutableDictionary (MSIDExtensions)
 
 - (void)msidSetObjectIfNotNil:(id)dictObject forKey:(id<NSCopying>)dictKey;
+- (void)msidSetNonEmptyString:(NSString *)stringObj forKey:(id<NSCopying>)dictKey;
```
unpluggedk: Q: do we really need this to be different to the above? I think we can simply modify above method to achieve what we want.

PR #311, opened 2018-12-03
fix a couple of compiler warnings

d-col:


d-col:
Separately, I'll look at the Xcode project settings.

PR #334, opened 2019-01-03
Changes needed for cache integration

jasoncoolmax:


IdentityCore/src/cache/accessor/MSIDDefaultTokenCacheAccessor.m
```diff
@@ -372,6 +372,24 @@ - (BOOL)removeAccessToken:(MSIDAccessToken *)token
             [MSIDTelemetry stopCacheEvent:event withItem:nil success:NO context:context];
             return nil;
         }
+#if AD_BROKER
```
antrix1989: Since this logic will be used only in the broker, maybe it is better to subclass MSIDDefaultTokenCacheAccessor and override this method so we can get rid of "#if ..."?

unpluggedk: I am fine with #if AD_BROKER, but content within this, we could extract as a method for better readability/debuggability esp. for non broker scenarios.

jasoncoolmax: addressed. Moved the logic back to broker :)

PR #336, opened 2019-01-08
Broker related changes

antrix1989:


IdentityCore/src/configuration/webview/MSIDWebviewConfiguration.h
```diff
@@ -27,6 +27,10 @@
 
 #import <Foundation/Foundation.h>
 
+#if TARGET_OS_IPHONE
+#import <UIKit/UIKit.h>
+#endif
+
```
jasoncoolmax: Why do we need this? :)

antrix1989: because of these 2 properties:

```
#if TARGET_OS_IPHONE
@property (weak) UIViewController *parentController;
@property (readwrite) UIModalPresentationStyle presentationType;
#endif
```

We need to include UIModalPresentationStyle & UIViewController from uikit.

PR #337, opened 2019-01-11
Use MSSTS rather than AADV1 in MSIDAccountCacheItem unit tests.

d-col:


d-col:
This is no longer needed

PR #346, opened 2019-01-28
Add missing CPU architectures

ElektrojungeMS:
The common authentication library is missing `armv7s` and `arm64e` CPU architectures. This PR adds those libraries to unblock https://github.com/AzureAD/microsoft-authentication-library-for-objc/pull/443 which does the same thing for the MSAL library.

Let me know if you need more context and don't hesitate to get in touch.

Cheers!
Benjamin

### DIFF
--- a/IdentityCore/IdentityCore.xcodeproj/project.pbxproj
+++ b/IdentityCore/IdentityCore.xcodeproj/project.pbxproj
@@ -4055,6 +4055,11 @@
 			isa = XCBuildConfiguration;
 			baseConfigurationReference = D6CF4E981FC3626A00CD70C5 /* identitycore__idlib__ios.xcconfig */;
 			buildSettings = {
+				ARCHS = (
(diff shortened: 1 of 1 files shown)


ElektrojungeMS:
Hey @oldalton,

Yes, `arm64e` is currently in preview. We added it to the App Center SDK a bit ahead of time, the reason being that we want to stay ahead of the curve.

```App Store submission with this architecture won't be accepted. Although it should be automatically removed when distributing through Organizer, it might break App Store submissions if Organizer is not used.```

This should not be an issue as the Xcode Organizer as well as `xcodebuild` will strip away the architecture when you build an app. Distributing an app with an `arm64e` slice is very very tricky today and, as you said, neither TestFlight nor iTunes Connect accept the slice.

Let me know what you think and I'm happy to remove the arm64e slice temporarily (if necessary). Feel free to get in touch through Teams if you want to discuss in private. =)

PR #349, opened 2019-02-02
Support pkeyauth for network request

jasoncoolmax:
Added PKeyAuth support for network request.

Question:
1) Currently B2C network request will also have `x-ms-PkeyAuth=1.0` in header, will that be an issue?
2) Does Mac needs PKeyAuth support for token endpoint? If not, needs to change not to send `x-ms-PkeyAuth=1.0` for Mac.

Manual testing has been done to verify the flow for broker.

TODO:
Some unit testing might be needed for `MSIDPKeyAuthHandler.m`, which I don't have time to finish today :P

### DIFF
--- a/IdentityCore/src/webview/embeddedWebview/challangeHandlers/MSIDPKeyAuthHandler.m
+++ b/IdentityCore/src/webview/embeddedWebview/challangeHandlers/MSIDPKeyAuthHandler.m
@@ -86,4 +86,166 @@
     return YES;
 }
 
++ (void)handleWwwAuthenticateHeader:(NSString *)wwwAuthHeaderValue
+                         requestUrl:(NSURL *)requestUrl
+                            context:(id<MSIDRequestContext>)context
+                  completionHandler:(void (^)(NSString *authHeader, NSError *error))completionHandler
+{
+    //pkeyauth word length=8 + 1 whitespace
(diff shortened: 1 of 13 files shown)


IdentityCore/src/webview/embeddedWebview/challangeHandlers/MSIDPKeyAuthHandler.m
```diff
@@ -86,4 +86,166 @@ + (BOOL)handleChallenge:(NSString *)challengeUrl
...
+
+// Decodes the parameters that come in the Authorization header. We expect them in the following
+// format:
+//
+// <key>="<value>", key="<value>", key="<value>"
+// i.e. version="1.0",CertAuthorities="OU=MyOrganization,CN=MyThingy,DN=windows,DN=net",Context="context!"
+//
+// This parser is lenient on whitespace, and on the presence of enclosing quotation marks. It also
+// will allow commented out quotation marks
++ (NSDictionary *)parseAuthHeader:(NSString *)authHeader
```
jasoncoolmax: Didn’t have time to revisit this function yet, I remember last time we had some issue with the auth header if version= is not the first pair in the auth header. Need to check if it has been fixed in this function :P

PR #378, opened 2019-02-22
Url session queue

antrix1989:


unpluggedk:
Without specifying this, we are relying on GCD to correctly set QoS for us; expecting to see userInitiated or utility.
1. However, I observed cases in which QoS was set unspecified
2. It is an auth library, I'd expect our calls to be returned immediately to few seconds, not minutes.

Thus, I'm setting it userInitiated.

PR #381, opened 2019-03-01
Remove utid check in default token validator.

unpluggedk:


unpluggedk:
Might take a different approach - don't review this

unpluggedk:
Other account validator is done in #384

PR #408, opened 2019-03-28
add conditional check for kerberos challenge.

unpluggedk:
Skb for mac has a Kerberos issue, and currently, out of options on fixing kerberos cert issue from the customer side.
So, mitigation for now is to have ADAL support "disabling" kerberos challenge.
This PR address that.


unpluggedk:
ADAL 2.7.x does not use this portion of code. 
My mistake, and thus will remove this PR and branch.

PR #436, opened 2019-04-25
additionalAccountFields fix

d-col:
initWithJSONDictionary isn't setting additionalAccountFields


d-col:
@GTValentine The validation of the behavior for multiple threads hitting the cache that's in an upcoming PR uncovered that additionalAccountFields wasn't getting set. Alternatively, I can work around it in the tests if we don't want to do this. But it seems like a defect not to set additionalAccountFields when deserializing an account from the cache.

d-col:
Will discuss further w/ @oldalton  & @GTValentine off-line.

PR #618, opened 2019-11-13
Fix broker build (SSO branch)

antrix1989:


IdentityCore/src/controllers/broker/MSIDBrokerInteractiveController.m
```diff
@@ -129,6 +129,7 @@ - (void)acquireToken:(MSIDRequestCompletionBlock)completionBlock
 + (BOOL)canPerformRequest:(MSIDInteractiveRequestParameters *)requestParameters
 {
 #if AD_BROKER
+    (void)requestParameters;
```
jasoncoolmax: shall we use `#pragma unused(requestParameters)` as the same style is used elsewhere?

antrix1989: done

PR #654, opened 2019-12-19
Fix memory leak in Authentication/Safari services

rohitnarula7176:


IdentityCore/src/webview/systemWebview/ios/MSIDSFAuthenticationSessionHandler.m
```diff
@@ -66,10 +66,12 @@ - (void)startWithCompletionHandler:(MSIDWebUICompletionHandler)completionHandler
         {
             NSError *cancelledError = MSIDCreateError(MSIDErrorDomain, MSIDErrorUserCancel, @"User cancelled the authorization session.", nil, nil, nil, nil, nil, YES);
             
+            self.webAuthSession = nil;
             if (completionHandler) completionHandler(nil, cancelledError);
             return;
         }
         
+        self.webAuthSession = nil;
         completionHandler(callbackURL, authError);
```
jasoncoolmax: Do we want to nil out `webAuthSession` before line 62 too?

IdentityCore/src/webview/systemWebview/session/MSIDASWebAuthenticationSessionHandler.m
```diff
@@ -82,10 +82,12 @@ - (void)startWithCompletionHandler:(MSIDWebUICompletionHandler)completionHandler
         {
             NSError *cancelledError = MSIDCreateError(MSIDErrorDomain, MSIDErrorUserCancel, @"User cancelled the authorization session.", nil, nil, nil, nil, nil, YES);
             
+            self.webAuthSession = nil;
             if (completionHandler) completionHandler(nil, cancelledError);
             return;
         }
         
+        self.webAuthSession = nil;
```
jasoncoolmax: Do we want to nil out `webAuthSession` before line 76 too?

rohitnarula7176: Good point. We should. Addressed

IdentityCore/src/webview/systemWebview/ios/MSIDSFAuthenticationSessionHandler.m
rohitnarula7176: We should. Addressed.

PR #760, opened 2020-06-12
fix a failure unit test when running on 10.15

kaisong1990:
## Proposed changes

The change is to remove items in memory and keep those items in local cache(storage)

This PR is to address a unit test failure on 10.15 and above where we should not be able to get local cached items. However, the logic of query items is a combination of getting items from both local and memory which fails the unit test. This PR will help to remove the items in memory only to prove we cannot get items from local cache

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [x] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/tests/util/MSIDApplicationTestUtil.m
+++ b/IdentityCore/tests/util/MSIDApplicationTestUtil.m
@@ -61,22 +61,23 @@ static NSArray *s_canOpenURLSchemes = nil;
 #pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
 @implementation UIApplication (TestOverride)
 
-+ (void)load {
++ (void)load
+{
     
     Method openUrlStr = class_getInstanceMethod(UIApplication.class, @selector(openURL:));
-    Method To_openUrlStr = class_getInstanceMethod(UIApplication.class, @selector(TO_openURL:));
-    method_exchangeImplementations(openUrlStr, To_openUrlStr);
+    Method msidOpenUrlStr = class_getInstanceMethod(UIApplication.class, @selector(msidOpenURL:));
+    method_exchangeImplementations(openUrlStr, msidOpenUrlStr);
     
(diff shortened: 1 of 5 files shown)


kaisong1990:
@rohitnarula7176 @oldalton Thanks for approving, when I upgrade to use xcode 10.15 in Travis, I got test failures on iOS target due to methods cannot be overwritten anymore. I used the swizzling way and looks like the issue got resolved. Would like to borrow your eyes to make sure this won't break anything else

IdentityCore/tests/util/MSIDApplicationTestUtil.m
```diff
@@ -60,7 +61,22 @@ @interface UIApplication (TestOverride)
 #pragma clang diagnostic ignored "-Wobjc-protocol-method-implementation"
 @implementation UIApplication (TestOverride)
 
-- (BOOL)openURL:(NSURL *)url
++ (void)load {
```
antrix1989, line 64: nit: new line

kaisong1990: Good catch. Still in Swift style

IdentityCore/tests/util/MSIDApplicationTestUtil.m
```diff
@@ -70,7 +86,7 @@ - (BOOL)openURL:(NSURL *)url
     return s_onOpenUrl(url, nil);
 }
 
-- (BOOL)canOpenURL:(NSURL *)url
+- (BOOL)TO_canOpenURL:(NSURL *)url
```
antrix1989, line 89: Rename `TO_canOpenURL` to "msidCanOpenURL". According to our codestyle, we always start with small letter and don't use underscore.

kaisong1990: Will update

PR #799, opened 2020-07-22
Fix a test bug in MSIDMacKeychainTokenCacheTests.m

jbzdarkid:
## Proposed changes

Fix a test bug in MSIDMacKeychainTokenCacheTests.m where, if the loginKeychainEmpty key is not reset, then the test will fail, since `_dataSource` will be nil.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [x] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/tests/mac/MSIDMacKeychainTokenCacheTests.m
+++ b/IdentityCore/tests/mac/MSIDMacKeychainTokenCacheTests.m
@@ -83,7 +83,11 @@
     NSError *error;
     MSIDKeychainUtil *keychainUtil = [MSIDKeychainUtil sharedInstance];
     keychainUtil.teamId = @"FakeTeamId";
+    // To ensure that the _dataSource is initialized, we need mark the loginKeychain as non-empty.
+    // MSIDMacKeychainTokenCache::kLoginKeychainEmpty = @"LoginKeychainEmpty"
+    [[NSUserDefaults standardUserDefaults] setBool:NO forKey:@"LoginKeychainEmpty"];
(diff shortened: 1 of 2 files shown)


IdentityCore/tests/mac/MSIDMacKeychainTokenCacheTests.m
```diff
@@ -82,7 +82,11 @@ - (void)setUp
     NSError *error;
     MSIDKeychainUtil *keychainUtil = [MSIDKeychainUtil sharedInstance];
     keychainUtil.teamId = @"FakeTeamId";
-    _dataSource = [MSIDMacKeychainTokenCache new];
+    // To ensure that the _dataSource is initialized, we need mark the loginKeychain as non-empty.
+    // MSIDMacKeychainTokenCache::kLoginKeychainEmpty = @"LoginKeychainEmpty"
+    [[NSUserDefaults standardUserDefaults] setBool:NO forKey:@"LoginKeychainEmpty"];
+    _dataSource = [MSAIMSIDMacKeychainTokenCache new];
+    XCTAssertNotNil(_dataSource);
```
antrix1989, line 89: I don't think we should XCTAssert in "setup". This kind of asserts are supposed to be used inside tests. I think we should use "assert" instead, so if setup incorrect we will crash the test host.

jbzdarkid: We're already usinhg XCTAssert in setup, see L112.
I'm happy to change it, if you like.

PR #893, opened 2020-11-11
ADAL fix for ADFS PKeyAuth issue. 

NerevarineRule:
## Proposed changes

When PKeyAuth is enabled through UA string, Client-side PKeyAuth handler creates a PkeyAuth header and then responds to the server-side challenge.

There was a bug in the client-side code, specifically in the msidEncodeStringFromDictionary API in the MSID category method for the NSString object.

This API constructs a URL string from the key-value pairs contained in the input dictionary.
ex: https://fs.msidlab4.com/adfs/ls/?&key1=value1&key2=value2&......&keyn=valuen.

This input dictionary is constructed from the submitURL parameter contained within the PKeyAuth request URL returned by the server. this submitURL field contains the information about the endpoint to which the PKeyAuth response must be submitted.

This is how the flow goes:

Client submits PKeyAuth request through the UA string
Server sees the request, and returns the challenge request URL, upon which the client must respond with a valid PKeyAuth header to the endpoint stored in the challenge request URL's submitURL parameter.
Client first converts the submitURL field into a dictionary with different key value pairs, where each key-value is generated from the components separated by & field, and each component takes the following format: &key1=value1&key2=value2....&keyn=valuen.
Then, client adds the "x-client-Ver" key-value pair to the dictionary in the step 3.
Client converts the dictionary back into the URL format using msidEncodeStringFromDictionary API.
During the transition from liveID endpoint to OrgID endpoint, namely, between AAD and ADFS transition point, AAD's JS logic appends three additional parameters: cbctx, lc, and mkt to the ADFS endpoint URL string, and this also propagates to the SubmitURL field in the PKeyAuth request URL string. However, they do not contain any additional information, so the SubmitURL would have the following structure: &cbctx=&lc=&mkt=

When they are converted into key-value pairs & stored in a dictionary, this dictionary is then fed into msidEncodeStringFromDictionary API after "x-client-Ver" key-value pair is added.

However, this API does not add "=" after the key, if the value a non-null empty string.
So, the end result is a URL that will have &cbctx&lc&mkt in the string.
This breaks ADFS's input parser logic, since it also delimits the URL string based on '=' character, when not having that character would throw an exception.

The fix is to add "=" in the WWWFormURLEncode API, even if the value is an empty string, and this complies with the URL query string spec as well, and both AAD and ADFS also comply with this convention.

Additionally, ADAL first extracts certificate issuer from the WPJ keychain, and then tries to look for it in the list of certificate authorities embedded in the challenge data returned by the server using regex matching. That list of certificate authorities is stored in the string format, but it was not decoded properly in ADAL's latest release branch/master, while MSAL does. 

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/util/NSString+MSIDExtensions.m
+++ b/IdentityCore/src/util/NSString+MSIDExtensions.m
@@ -94,7 +94,6 @@ typedef unsigned char byte;
     return [self stringByTrimmingCharactersInSet:set];
 }
 
-
 - (NSString *)msidWWWFormURLDecode
 {
     // Two step decode: first replace + with a space, then percent unescape
@@ -109,6 +108,10 @@ typedef unsigned char byte;
     return CFBridgingRelease(unescapedString);
 }
 
+- (NSString *)msidURLDecode
+{
+    return CFBridgingRelease(CFURLCreateStringByReplacingPercentEscapes(kCFAllocatorDefault, (CFStringRef)self, CFSTR("")));
+}
 
 - (NSString *)msidWWWFormURLEncode
 {
@@ -224,6 +227,11 @@ typedef unsigned char byte;
          NSString *encodedValue = [[v msidTrimmedString] msidWWWFormURLEncode];
          
          if (![NSString msidIsStringNilOrBlank:encodedValue])
+         {
+             [encodedString appendFormat:@"=%@", encodedValue];
+         }
+        
+         else if ((encodedValue && ![encodedValue isKindOfClass:[NSNull class]]) && !encodedValue.length)
          {
              [encodedString appendFormat:@"=%@", encodedValue];
          }
--- a/IdentityCore/src/webview/embeddedWebview/ui/ios/MSIDWebviewUIController.m
+++ b/IdentityCore/src/webview/embeddedWebview/ui/ios/MSIDWebviewUIController.m
@@ -58,6 +58,11 @@ static WKWebViewConfiguration *s_webConfig;
 {
     WKWebViewConfiguration *webConfig = [WKWebViewConfiguration new];
 
+    if (@available(iOS 9.0, *))
+    {
+        webConfig.applicationNameForUserAgent = kMSIDPKeyAuthKeyWordForUserAgent;
+    }
+    
     if (@available(iOS 13.0, *))
     {
         webConfig.defaultWebpagePreferences.preferredContentMode = WKContentModeMobile;
--- a/IdentityCore/src/workplacejoin/MSIDPkeyAuthHelper.m
+++ b/IdentityCore/src/workplacejoin/MSIDPkeyAuthHelper.m
@@ -110,7 +110,7 @@
 {
     NSString *regexString = @"OU=[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}";
     keychainCertIssuer = [keychainCertIssuer uppercaseString];
-    certAuths = [certAuths uppercaseString];
+    certAuths = [certAuths.msidURLDecode uppercaseString];
     NSRegularExpression *regex = [NSRegularExpression regularExpressionWithPattern:regexString options:0 error:NULL];
     
     for (NSTextCheckingResult *myMatch in [regex matchesInString:certAuths options:0 range:NSMakeRange(0, [certAuths length])]){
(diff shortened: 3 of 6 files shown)


IdentityCore/src/util/NSString+MSIDExtensions.m
```diff
@@ -227,6 +230,11 @@ + (NSString *)msidWWWFormURLEncodedStringFromDictionary:(NSDictionary *)dict
          {
              [encodedString appendFormat:@"=%@", encodedValue];
          }
+        
+         else if ((encodedValue && ![encodedValue isKindOfClass:[NSNull class]]) && !encodedValue.length)
```
ameyapat, line 234: if encodedValue is nil, this logic will not attach it to encodedString. Is that okay? Making encodedValue = @"" if it is nilOrBlank might be useful?

IdentityCore/src/workplacejoin/MSIDPkeyAuthHelper.m
```diff
@@ -110,7 +110,7 @@ + (BOOL)isValidIssuer:(NSString *)certAuths
 {
     NSString *regexString = @"OU=[a-fA-F0-9]{8}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{4}-[a-fA-F0-9]{12}";
     keychainCertIssuer = [keychainCertIssuer uppercaseString];
-    certAuths = [certAuths uppercaseString];
+    certAuths = [certAuths.msidURLDecode uppercaseString];
```
ameyapat, line 113: Is msidURLDecode different from msidWWWFormURLDecode? I can see msidWWWFormURLDecode replacing + with a space first

IdentityCore/src/util/NSString+MSIDExtensions.m
NerevarineRule: Yeah good point. I think we should append "=" even for nil or blank scenario. I believe that calling APIs normally convert URLs first into a dictionary with multiple key-value pairs from &key1=value1 substrs and then add some additional key-value pairs and then convert it back to a new URL. I assumed that the check for NULL or blank would be redundant, as that might never happen was my understanding.

I think we should either 
1) create an error for nilorblank, as that might indicate an error in the calling API's logic that converted the original URL into a query dictionary
2) Simply add "=" to the keyword regardless. 

Thanks for your suggestion.

IdentityCore/src/workplacejoin/MSIDPkeyAuthHelper.m
NerevarineRule: Good point. Thanks for bringing this up. However, they are different.
EX:
Input: 
OU%3d5ce5e2c6-f36e-4e53-840b-98acf367a25e%2cCN%3dMS-Organization-Access+%2cDC%3dMSIDLAB4+%2cDC%3dcom+%3bOU%3d82dbaca4-3e81-46ca-9c73-0950c1eaca97%2cCN%3dMS-Organization-Access+%2cDC%3dwindows+%2cDC%3dnet+

msidURLDecode output:
OU=5CE5E2C6-F36E-4E53-840B-98ACF367A25E,CN=MS-ORGANIZATION-ACCESS+,DC=MSIDLAB4+,DC=COM+;OU=82DBACA4-3E81-46CA-9C73-0950C1EACA97,CN=MS-ORGANIZATION-ACCESS+,DC=WINDOWS+,DC=NET+

msidWWWFormURLDecode:
OU=5CE5E2C6-F36E-4E53-840B-98ACF367A25E,CN=MS-ORGANIZATION-ACCESS ,DC=MSIDLAB4 ,DC=COM ;OU=82DBACA4-3E81-46CA-9C73-0950C1EACA97,CN=MS-ORGANIZATION-ACCESS ,DC=WINDOWS ,DC=NET

But yep. Technically, they both will work, since the subsequent regex matching will be able to grep the correct substr out of both cases.

I just decided to comply with how the latest MSAL dev branch handles both URL encode & decode, as they are more spec-compliant based on this PR - https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/350

IdentityCore/src/webview/embeddedWebview/ui/ios/MSIDWebviewUIController.m
```diff
@@ -58,6 +58,11 @@ + (WKWebViewConfiguration *)defaultWKWebviewConfiguration
 {
     WKWebViewConfiguration *webConfig = [WKWebViewConfiguration new];
 
+    if (@available(iOS 9.0, *))
```
kaisong1990, line 61: Are we still supporting iOS 9?

NerevarineRule: good catch. This was just to follow the starting version that supports this API. the end result should be the same, even though our min supported version is iOS 11.

PR #922, opened 2021-01-07
Generic Thread-safe MSIDLRUCache

NerevarineRule:
## Proposed changes

While implementing throttling cache, it was brought up that it might be useful to extend the class to be more generic and usable by other data types. 

Here's the implementation for generic LRU cache + unit test coverage.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+
+- (instancetype)initWithCacheSize:(NSUInteger)cacheSize;
+
++ (MSIDLRUCache *)sharedInstance;
+
+/**
+add new node to the front of LRU cache.
+if node already exists, update and move it to the front of LRU cache
+ */
+- (BOOL)addToCache:(KeyType)key
```
antrix1989, line 43: nit: might be better to name it: `setCacheRecord:forKey:error`

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+       cacheRecord:(nullable ObjectType)cacheRecord
+             error:(NSError *__nullable*__nullable)error;
+
+- (BOOL)removeFromCache:(KeyType)key
```
antrix1989, line 47: nit: might be better to name it: removeCacheRecordForKey:error

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+                  error:(NSError *__nullable*__nullable)error;
+
+/**
+ retrieve cache record from the corresponding node, and move the node to the front of LRU cache.
+ Additionally, pass in selector that can be used as criteria to remove records
+ */
+- (nullable ObjectType)updateAndReturnCacheRecord:(KeyType)key
```
antrix1989, line 54: nit: should we name it `objectForKey:`?

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+                                            error:(NSError *__nullable*__nullable)error;
+
+- (nullable NSArray<ObjectType> *)enumerateAndReturnAllObjects; //return all cached elements without disturbing order
```
antrix1989, line 57: nit: should we name it `allObjects`?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+
+//Main class 
+@interface MSIDLRUCache ()
+
+@property (nonatomic) NSUInteger cacheSizeInt;
+@property (nonatomic) NSUInteger cacheUpdateCountInt;
+@property (nonatomic) MSIDLRUCacheNode *head;
+@property (nonatomic) MSIDLRUCacheNode *tail;
+@property (nonatomic) NSMutableDictionary *container;
+@property (nonatomic) NSMutableDictionary *key_signature_map;
```
antrix1989, line 81: Followup our code style, should be renamed to keySignatureMap

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+@property (nonatomic) dispatch_queue_t synchronizationQueue;
+
+@end
+
+@implementation MSIDLRUCache
+
+- (NSUInteger)cacheSize
+{
+    return self.cacheSizeInt-2;
```
antrix1989, line 90: What is '2' here? Looks like a magic number. Can you move it to constant?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+- (BOOL)addToCache:(id)key
+       cacheRecord:(id)cacheRecord
+             error:(NSError *__nullable*__nullable)error
+{
+    __block NSError *subError = nil;
+    BOOL result = YES;
+    dispatch_barrier_sync(self.synchronizationQueue, ^{
+        if (self.cacheSizeInt <= 2)
+        {
+            subError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal, @"MSIDLRUCache Error: Attempting to write to an empty cache!", nil, nil, nil, nil, nil, YES);
```
antrix1989, line 156: The error message is a bit confusing. I think you should validate the provided cache size in the init, so it should be >= 1. Then you can remove validation here.

antrix1989:
Added few minor comment, otherwise looks good. And thanks for the comments in the code, they help!

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+    if (![self.container objectForKey:signature])
+    {
+        if (error)
+        {
+            *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal, @"MSIDLRUCache Error: Unable to find valid node for the input signature during removal", nil, nil, nil, nil, nil, YES);
+        }
+        return NO;
+    }
+    
+    MSIDLRUCacheNode *node = [self.container objectForKey:signature];
```
jasoncoolmax, line 245: can we make some optimization together with line 236?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+    
+    /**
+    BEFORE:
+     A->B->C
+     A<-B<-C
+    AFTER:
+     A->C
+     A<-C
+     */
+    currentHeadNode.prevSignature = [node.signature mutableCopy];
```
jasoncoolmax, line 332: any reason why we need to do `mutableCopy`?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+    MSIDLRUCacheNode *currentHeadNode = [self.container objectForKey:currentHeadSignature];
+    
+    /**
+    BEFORE:
+     A->B->C
+     A<-B<-C
+    AFTER:
+     A->C
+     A<-C
+     */
```
jasoncoolmax, line 331: the comment part is referring to removal?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+            MSIDLRUCacheNode *node = [self.container objectForKey:signature];
+            [signature setString:node.nextSignature];
+            [res addObject:node];
+        }
+    }
+    return res;
+}
+
+- (NSString *)generateRandomSignature //mock pointer 62^8 ~ 2^14 randomness
+{
```
jasoncoolmax, line 412: 2^14 = 16384. Is it correct that: if this function generates duplicate numbers, the linked list you constructed in this class will become a graph? (any reason not to use NSUUID?)

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+    NSMutableArray *res;
+    if (![self.head.nextSignature isEqualToString:TAIL_SIGNATURE])
+    {
+        res = [NSMutableArray new];
+        NSMutableString *signature = [self.head.nextSignature mutableCopy];
+        
+        while (![signature isEqualToString:TAIL_SIGNATURE])
+        {
+            MSIDLRUCacheNode *node = [self.container objectForKey:signature];
+            [signature setString:node.nextSignature];
```
jasoncoolmax, line 386: just curious, what is the reason that we don't use `signature = node.nextSignature`? :)

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+            [res addObject:node.cacheRecord];
```
jasoncoolmax, line 387: will a nil check here prevent us from crash? :)

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: yep.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: input param is NSString, and this variable is NSMutable string.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+                    nextSignature:(NSString *)nextSignature
+                      cacheRecord:(id)cacheRecord
+{
+    self = [super init];
+    if (self)
+    {
+        _signature = signature;
+        _prevSignature = [prevSignature mutableCopy];
+        _nextSignature = [nextSignature mutableCopy];
+        _cacheRecord = cacheRecord;
```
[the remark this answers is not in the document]
NerevarineRule: those that use mutableCopy are assigning NSString to NSMutableString, while other variables don't need such conversion.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: sorry there's a typo in the comment. this is actually 2*10^14. sorry for the confusion. 
NSUUID also works :) just wanted to have more control over the length and randomness of the pseudo-pointer.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: It's to suppress warning when assigning NSString type to NSMutable String. Incompatible pointer assignment.

IdentityCore/src/requests/MSIDSilentTokenRequest.m
```diff
@@ -419,6 +419,7 @@ - (void)acquireTokenWithRefreshTokenImpl:(MSIDBaseToken<MSIDRefreshableToken> *)
     MSIDRefreshTokenGrantRequest *tokenRequest = [self.oauthFactory refreshTokenRequestWithRequestParameters:self.requestParameters
                                                                                                 refreshToken:refreshToken.refreshToken];
 
+    //Initialize wrapper request object to store request parameters required
```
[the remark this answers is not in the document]
hieunguyenmsft: +1, should remove it as it's not related to this PR

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+
+
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSIDLRUCache <KeyType, ObjectType>: NSObject
+
+@property (nonatomic, readonly) NSUInteger cacheSize; //size of the LRU cache
+@property (nonatomic, readonly) NSUInteger numCacheRecords; //number of valid records currently stored in the LRU cache
+@property (nonatomic, readonly) NSUInteger cacheUpdateCount; //number of times cache entry has been updated
```
hieunguyenmsft, line 33: Just curious, what's the use for the variable `cacheUpdateCount`? If we want to count each time a cache updated, do we want to count each time entry is evicted?

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+- (BOOL)addToCache:(KeyType)key
+       cacheRecord:(nullable ObjectType)cacheRecord
+             error:(NSError *__nullable*__nullable)error;
+
+- (BOOL)removeFromCache:(KeyType)key
+                  error:(NSError *__nullable*__nullable)error;
+
+/**
+ retrieve cache record from the corresponding node, and move the node to the front of LRU cache.
+ Additionally, pass in selector that can be used as criteria to remove records
```
hieunguyenmsft, line 52: Can you please explain the line: 
` Additionally, pass in selector that can be used as criteria to remove records`?

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
+ */
+- (nullable ObjectType)updateAndReturnCacheRecord:(KeyType)key
+                                            error:(NSError *__nullable*__nullable)error;
+
+- (nullable NSArray<ObjectType> *)enumerateAndReturnAllObjects; //return all cached elements without disturbing order
+
```
hieunguyenmsft, line 58: Should we provide an API to clear all the current records in cache, for example `clearAll`?

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: good catch. I should remove nullable flag from addToCache param, by doing so it will enforce non-null input into addToCache param, and by doing so, enumerate API will also ensure no such case arises.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,432 @@
...
+    }
+    return self;
+}
+
+@end
```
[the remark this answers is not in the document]
NerevarineRule: Since these are private helper classes not used by anyone else, I thought it best to keep it in the same file. Also, by doing so, I can avoid having mutable public properties.

IdentityCore/src/throttling/cache/MSIDLRUCache.h
hieunguyenmsft, line 57: A quick note about the order of the return elements would bring more benefit :) I think in this case you return in ordered from most recently accessed to least recently accessed, right?

IdentityCore/src/throttling/cache/MSIDLRUCache.h
```diff
@@ -0,0 +1,61 @@
...
++ (MSIDLRUCache *)sharedInstance;
+
+/**
+add new node to the front of LRU cache.
+if node already exists, update and move it to the front of LRU cache
+ */
+- (BOOL)addToCache:(KeyType)key
+       cacheRecord:(nullable ObjectType)cacheRecord
+             error:(NSError *__nullable*__nullable)error;
```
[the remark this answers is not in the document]
NerevarineRule: sounds good.

hieunguyenmsft (commented):
Some comments in the header file. I haven't looked at the implementation it, will check it later :)

IdentityCore/src/requests/MSIDSilentTokenRequest.m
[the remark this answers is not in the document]
NerevarineRule: This was dummy comment to indicate where to place throttling logic. Thanks for catching it. removed :)

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: sounds good.

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: we want to indicate that it also moves the node to front. agreed the naming is a bit confusing. I will rename it to addOrMoveToFront

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: good catch.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: we insert placeholder head and tail node during initialization for easier handling of node removal.

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: For this, I think i will leave the naming as is to differentiate between node vs object.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
NerevarineRule: Yeah the message is confusing. I will change it to. Cache was initialized as an empty cache. Cannot write to cache of size 0.

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: yeah good call! Let's add cacheEvictionCount also.

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: Good catch. I was playing around with different capabilities and decided to scrap it, but forgot to remove the comment. Thanks :)

IdentityCore/src/throttling/cache/MSIDLRUCache.h
NerevarineRule: Yeah good idea.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,449 @@
...
+@property (nonatomic) NSMutableDictionary *keySignatureMap;
+@property (nonatomic) dispatch_queue_t synchronizationQueue;
+
+@end
+
+@implementation MSIDLRUCache
+
+- (NSUInteger)cacheSize
+{
+    return self.cacheSizeInt-DEFAULT_CACHE_OFFSET_SIZE;
```
antrix1989, line 88: nit: spaces around "-"

IdentityCore/src/throttling/cache/MSIDLRUCache.m
```diff
@@ -0,0 +1,449 @@
...
+}
+
+- (NSUInteger)numCacheRecords
+{
+    return self.container.allKeys.count-DEFAULT_CACHE_OFFSET_SIZE;
```
antrix1989, line 93: nit: spaces

IdentityCore/src/throttling/cache/MSIDLRUCache.m
antrix1989: +1 to use NSUUID here. We can simplify the logic here and reduce complexity.

IdentityCore/src/throttling/cache/MSIDLRUCache.m
antrix1989: > by doing so it will enforce non-null input

It is still possible to pass null even if param is declared as "no-null". It safe to check for nil before you add to the array.

antrix1989 (approved):
Approved with few comments.

PR #923, opened 2021-01-11
Extend backgroud tasks for silent and interactive request

kaisong1990:
## Issue trying to resolve

We have customers reporting if they have MFA and put Authenticator app into background for 2FA, they are getting "The network connection was lost." or "The Internet connection appears to be offline." error from Authenticator app and cannot finish the whole authentication. 

When putting app into background, for interactive request, app will finish the authcode process first and then end the background task. At that moment, the app will be suspended. When reopen the app after 2nd FA is done, the AuthCode-token exchange flow will continue, however server drops the connection (not sure why) already and caused the network error.

## Proposed changes

Instead of closing the background task when the web view (both ASWebSession and WKWebView ) is dismissed, we will stop the task once the whole interactive/silent request is finished either in result or error state.

## Q&A

Why we need to call/use background tasks?
When your app moves to the background, the system calls your app delegate’s applicationDidEnterBackground(_:) method. That method has five seconds to perform any tasks and return. Shortly after that method returns, the system puts your app into the suspended state. 

What if we don't add proper code to stop the background task?
For each task before time expires, the system kills the app. If you provide a block object in the handler parameter, the system calls your handler before time expires to give you a chance to end the task. Current we will stop the current task in the handler

What type of background tasks we have?
Currently there are two, one for interactive request, and another for silent request. They are isolated.

What happens if same background task already happening?
We will check if the current background task is already in the cache based on background task id, and if there was already, we would simply return instead of adding it again

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/controllers/MSIDSilentController.m
+++ b/IdentityCore/src/controllers/MSIDSilentController.m
@@ -29,6 +29,9 @@
 #import "MSIDTelemetryEventStrings.h"
 #import "MSIDTokenResult.h"
 #import "MSIDAccount.h"
+#if TARGET_OS_IPHONE
+#import "MSIDBackgroundTaskManager.h"
+#endif
 
 @interface MSIDSilentController()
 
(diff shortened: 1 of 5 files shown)


IdentityCore/src/controllers/MSIDSilentController.m
```diff
@@ -29,6 +29,9 @@
 #import "MSIDTelemetryEventStrings.h"
```
kaisong1990, line 29: This pull request does not update changelog.txt.

oldalton (approved):

hieunguyenmsft:
Do you think this problem starts at ios 14? I remember playing with the background testing several weeks ago and didn't see this error, so just curious when will it happen.

kaisong1990:
> Do you think this problem starts at ios 14? I remember playing with the background testing several weeks ago and didn't see this error, so just curious when will it happen.

@oldalton Helped to test on iOS 13 and didn't repro it, and I found there was customer reporting this issue since 14.2. User has to leave Authenticator for MFA for certain amount of time, so it might not be that straight to reproduce

PR #971, opened 2021-03-25
Update telemetry string size limit to ESTS's new limits

mipetriu:
## Proposed changes

ESTS has recently introduced a change to truncate telemetry strings in the "x-client-current-telemetry" and "x-client-last-telemetry" headers if they exceed 100 and 350 bytes respectively (the previous max limit for each string was 4kB). While we currently have nothing that will break if telemetry data is truncated on the server-side, it is useful to update the limits in our code so we are not sending excess data in our network requests. Also, we have code that sends telemetry data that has been cut off due to size during serialization in subsequent requests, so that can preserve the data that the server would truncate.

The last request serialization object will now override the telemetry string size methods in order to keep its own size limit value.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [x] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/telemetry/request_telemetry/MSIDCurrentRequestTelemetrySerializedItem.m
+++ b/IdentityCore/src/telemetry/request_telemetry/MSIDCurrentRequestTelemetrySerializedItem.m
@@ -64,7 +64,7 @@ static int telemetryStringSizeLimit = 100;
 {
     NSString *telemetryString = [NSString stringWithFormat:@"%@|%@|%@", self.schemaVersion, [self serializeFields: self.defaultFields], [self serializeFields: self.platformFields]];
     
-    if ([telemetryString lengthOfBytesUsingEncoding:NSUTF8StringEncoding] > telemetryStringSizeLimit)
+    if ((int)[telemetryString lengthOfBytesUsingEncoding:NSUTF8StringEncoding] > [MSIDCurrentRequestTelemetrySerializedItem telemetryStringSizeLimit])
     {
         return nil;
     }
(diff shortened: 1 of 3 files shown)


IdentityCore/src/telemetry/request_telemetry/MSIDCurrentRequestTelemetrySerializedItem.m
```diff
@@ -64,7 +64,7 @@ - (NSString *)serialize
 {
     NSString *telemetryString = [NSString stringWithFormat:@"%@|%@|%@", self.schemaVersion, [self serializeFields: self.defaultFields], [self serializeFields: self.platformFields]];
     
-    if ([telemetryString lengthOfBytesUsingEncoding:NSUTF8StringEncoding] > 4000)
+    if ([telemetryString lengthOfBytesUsingEncoding:NSUTF8StringEncoding] > telemetryStringSizeLimit)
```
Veena11, line 67: Build failure due to comparison type mismatch can be fixed by changing this line :
if ((int)([telemetryString lengthOfBytesUsingEncoding:NSUTF8StringEncoding]) > telemetryStringSizeLimit)

Veena11 (commented):
Code change looks good. Fix for the build failure is added in comment

IdentityCore/src/telemetry/request_telemetry/MSIDCurrentRequestTelemetrySerializedItem.m
mipetriu: Thanks for pointing it out, its addressed

oldalton (approved):

PR #1004, opened 2021-06-17
Add nil check for sourceApplication and do a early return if response…

kaisong1990:
… is not from broker

The comment mentioned should return early when sourcerApplication, but looks like the logic is missing. Add a fix in

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/controllers/broker/ios/MSIDBrokerInteractiveController.m
+++ b/IdentityCore/src/controllers/broker/ios/MSIDBrokerInteractiveController.m
@@ -288,7 +288,7 @@ static MSIDBrokerInteractiveController *s_currentExecutingController;
            sourceApplication:(nullable NSString *)sourceApplication
        brokerResponseHandler:(nonnull MSIDBrokerResponseHandler *)responseHandler
 {
-    // sourceApplication could be nil, we want to return early if we know for sure response is not from broker
+    // sourceApplication could be nil or empty, we want to return early if we know for sure response is not from broker
     if (![NSString msidIsStringNilOrBlank:sourceApplication] && ![self isResponseFromBroker:sourceApplication])
     {
         MSID_LOG_WITH_CTX(MSIDLogLevelWarning,nil, @"Asked to handle non broker response. Skipping request.");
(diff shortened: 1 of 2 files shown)


IdentityCore/src/controllers/broker/ios/MSIDBrokerInteractiveController.m
```diff
@@ -288,7 +288,7 @@ + (BOOL)completeAcquireToken:(nullable NSURL *)resultURL
        brokerResponseHandler:(nonnull MSIDBrokerResponseHandler *)responseHandler
```
kaisong1990, line 288: This pull request does not update changelog.txt.

IdentityCore/src/controllers/broker/ios/MSIDBrokerInteractiveController.m
```diff
@@ -288,7 +288,7 @@ + (BOOL)completeAcquireToken:(nullable NSURL *)resultURL
...
 {
     // sourceApplication could be nil, we want to return early if we know for sure response is not from broker
-    if (sourceApplication && ![self isResponseFromBroker:sourceApplication])
+    if (!sourceApplication || (sourceApplication && ![self isResponseFromBroker:sourceApplication]))
```
oldalton, line 291: How would this work if sourceApplication is nil? (This happens every time when app is non-Microsoft signed)

oldalton (changes requested):
Added a question to clarify.

IdentityCore/src/controllers/broker/ios/MSIDBrokerInteractiveController.m
kaisong1990: Thanks for clarifying, I was confused by the comment in the code. In this case, ```!sourceApplication``` is a valid case, and should be able to process with the rest correct?

kaisong1990: Updated the PR was we discuseed

oldalton (approved):

ameyapat:
Hi @kaisong1990, please add a dummy commit or update changelog to trigger newly added required status check '[Common Core] PR Validation'

PR #1032, opened 2021-09-03
fix throttling error handling of MSIDLRUCache + add more loggings.

NerevarineRule:
## Proposed changes

When a given request is has not yet been throttled, its thumbprint value would not be cached by MSIDLRUCache. when that scenario is detected, throttling logic should just log that the request is not cached / throttled yet, and hence there is no cache record. 

ThrottlingModelForIncomingRequest check currently checks as below - but if error code is MSIDErrorThrottleCacheNoRecord, it won't trigger the if conditional and instead trigger the else conditional and mark it as error:

```
      if (!error.code == MSIDErrorThrottleCacheNoRecord && !error.code == MSIDErrorThrottleCacheInvalidSignature)
      {
            MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context, @"Throttling: No record in throttle cache");
            error = nil;
      }  
      else
      {
           //error is triggered. 
      }

```

```
but it should be 
if (error.code == MSIDErrorThrottleCacheNoRecord || error.code == MSIDErrorThrottleCacheInvalidSignature)

instead,

```
In order for the error code to be correctly handled for the scenario where the request is not throttled/cached.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/throttling/model/MSIDThrottlingModelFactory.m
```diff
@@ -50,9 +50,10 @@ + (MSIDThrottlingModelBase *)throttlingModelForIncomingRequest:(id<MSIDThumbprin
                                                                                                    error:&error];
     if (error)
     {
-        if (error.code != MSIDErrorThrottleCacheNoRecord && error.code != MSIDErrorThrottleCacheInvalidSignature)
+        if (error.code == MSIDErrorThrottleCacheNoRecord || error.code == MSIDErrorThrottleCacheInvalidSignature)
         {
             MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context, @"Throttling: No record in throttle cache");
+            error = nil;
```
hieunguyenmsft, line 56: nit: error is local, doesn't need to set back to nil

hieunguyenmsft (approved):
looking goods, thanks for fixing it and improving log.

changelog.txt
```diff
@@ -13,8 +15,7 @@ Version 1.6.7
 Version 1.6.6
 * Fix telemetry enum value for refresh_in getting overwritten (#996)
 * Update automation for Xcode 12 UI
-* Support for Universal cache in Login Keychain on macOS 10.15+ in Developers (#1016)
```
hieunguyenmsft, line 16: why there is changes log update for older version

hieunguyenmsft (approved):

oldalton (approved):

PR #1038, opened 2021-09-21
Modifying background task manager to allow stopping task once there are no relying requests present

ameyapat:
## Proposed changes

In case the UI Application (Authenticator for broker) is put in background, the OS kills the network task and we get an error -1005 "Network connection lost". To prevent the OS from doing so we can call [beginbackgroundtaskwithname](https://developer.apple.com/documentation/uikit/uiapplication/1623051-beginbackgroundtaskwithname).

Current implementation allows creating these background tasks for 2 types : Silent & Interactive requests. We call start and stop these background tasks through WebUI controller, System WebView controller &  Interactive and Silent Token request objects. A situation may happen where an Interactive bg task is started by WebUI controller and through the flow an interactive token request is initialized. During initialization, the interactive token request may attempt to start an interactive Bg task but doesn't since one was already started by WebUi controller. But, once interactive token request completes and deallocates it calls background task manager to end interactive bg task that was started by the WebUi controller. WebUi controller loses network socket allocation since its task was prematurely cancelled. Similarly for silent flow.

As a fix, this PR aims to add logic so that starting and stopping background task operations will involve keeping track of a counter for each type of background task. The counter is incremented by 1 for a start bg task operation and decremented by 1 for a stop operation. The background task operation for a type is stopped only when this counter value reaches 0, this will ensure there are 0 requests of a particular type occurring when app is in bg and we can stop bg task operation.

This scenario can also extends to Broker requests which covers other type of requests like nonce, open id , discovery etc.
This PR also adds another type for broker requests. Same logic would apply for broker requests.

Note: The OS can expire a background task operation when the task is in background for too long. This time is typically 30 seconds. In the expiry handler we will stop the operation regardless of other requests relying on a background operation to be present.

I'll add Unit tests after any feedback.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
+++ b/IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
@@ -71,38 +71,41 @@
 
 - (void)startOperationWithType:(MSIDBackgroundTaskType)type
 {
-    MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
-    
-    if (backgroundTaskData)
+    @synchronized (self.taskCache)
     {
-        // A background task was already started for this type, updating count
-        backgroundTaskData.callerReferenceCount++;
-        return;
+        __weak MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
+        if (backgroundTaskData)
+        {
+            // A background task was already started for this type, updating count
+            backgroundTaskData.callerReferenceCount++;
+            return;
+        }
+        UIBackgroundTaskIdentifier backgroundTaskId = [[MSIDAppExtensionUtil sharedApplication] beginBackgroundTaskWithName:@"Interactive login"
+                                                                                                          expirationHandler:^{
+                                                                        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Background task expired for type %ld.", (long)type);
+                                                                    // If a task took too long & OS has decided to kill it, end bg task for that type regardless of other requsts relying on bg protection.
+                                                                        [self expireOperationWithType:type];
+                                                                    }];
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Start background app task with type %ld & taskId : %lu", (long)type, (unsigned long)backgroundTaskId);
+        [self setBackgroundTask:[[MSIDBackgroundTaskData alloc] initWithTaskId:backgroundTaskId] forType:type];
     }
-    
-    UIBackgroundTaskIdentifier backgroundTaskId = [[MSIDAppExtensionUtil sharedApplication] beginBackgroundTaskWithName:@"Interactive login"
-                                                                  expirationHandler:^{
-                                                                      MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Background task expired for type %ld.", (long)type);
-                                                                // If a task took too long & OS has decided to kill it, end bg task for that type regardless of other requsts relying on bg protection.
-                                                                      [self expireOperationWithType:type];
-                                                                  }];
-    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Start background app task with type %ld & taskId : %lu", (long)type, (unsigned long)backgroundTaskId);
-    [self setBackgroundTask:[[MSIDBackgroundTaskData alloc] initWithTaskId:backgroundTaskId] forType:type];
 }
 
 - (void)stopOperationWithType:(MSIDBackgroundTaskType)type
 {
-    MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
-    backgroundTaskData.callerReferenceCount--;
-    if (!backgroundTaskData || backgroundTaskData.callerReferenceCount > 0)
+    @synchronized (self.taskCache)
     {
-        // No background task found in task cache for specified type or there are still other tasks relying on background protection.
-        return;
+        __weak MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
+        backgroundTaskData.callerReferenceCount--;
+        if (!backgroundTaskData || backgroundTaskData.callerReferenceCount > 0)
+        {
+            // No background task found in task cache for specified type or there are still other tasks relying on background protection.
+            return;
+        }
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Stop background task with type %ld & taskId : %lu", (long)type, backgroundTaskData.backgroundTaskId);
+        [[MSIDAppExtensionUtil sharedApplication] endBackgroundTask:backgroundTaskData.backgroundTaskId];
+        [self.taskCache removeObjectForKey:@(type)];
     }
-    
-    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Stop background task with type %ld & taskId : %lu", (long)type, backgroundTaskData.backgroundTaskId);
-    [[MSIDAppExtensionUtil sharedApplication] endBackgroundTask:backgroundTaskData.backgroundTaskId];
-    [self.taskCache removeObjectForKey:@(type)];
 }
 
 - (void)expireOperationWithType:(MSIDBackgroundTaskType)type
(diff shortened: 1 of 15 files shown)


IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
```diff
@@ -68,52 +94,67 @@ + (MSIDBackgroundTaskManager *)sharedInstance
...
-    [self setBackgroundTask:UIBackgroundTaskInvalid forType:type];
+    [[MSIDAppExtensionUtil sharedApplication] endBackgroundTask:backgroundTaskData.backgroundTaskId];
+    [self setBackgroundTask:[[MSIDBackgroundTaskData alloc] initWithTaskId:UIBackgroundTaskInvalid caller:backgroundTaskData.callerReference] forType:type];
+}
+
+- (void)stopOperationWithType:(MSIDBackgroundTaskType)type caller:(nonnull id)caller
+{
+    MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
+    
+    if ((backgroundTaskData.backgroundTaskId == UIBackgroundTaskInvalid) || (![caller isEqual:backgroundTaskData.callerReference]))
```
kaisong1990, line 137: What does the ```isEqual``` refers to? Memory address?

kaisong1990, line 137: Previously we have different bg status for bg types. Does it make sense to simply remove the bg object if it is invalid, instead of setting an invalid bg object?

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
```diff
@@ -35,6 +35,32 @@ @interface MSIDBackgroundTaskManager()
...
+    if (self)
+    {
+        _backgroundTaskId = backgroundTaskId;
+        __typeof__(caller) __weak weakCallerReference = caller;
+        _callerReference = weakCallerReference;
+    }
+    return self;
+}
+@end
+
```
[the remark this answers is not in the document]
ameyapat: I would want to keep MSIDBackgroundTaskData to be used only in this file so added it here. Do you suggest creating a separate file and adding it as a private category?

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
```diff
@@ -35,6 +35,32 @@ @interface MSIDBackgroundTaskManager()
...
+@implementation MSIDBackgroundTaskData
+
+-(id)initWithTaskId:(UIBackgroundTaskIdentifier)backgroundTaskId caller:(id)caller
+{
+    self = [super init];
+    if (self)
+    {
+        _backgroundTaskId = backgroundTaskId;
+        __typeof__(caller) __weak weakCallerReference = caller;
```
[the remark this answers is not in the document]
ameyapat: I think a strong reference present in the cache might be dangerous. Although we can remove object from cache once stopOperation is called, currently we don't do so, to be extra careful I kept it as a weak ref instead of strong.

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
```diff
@@ -68,52 +94,67 @@ + (MSIDBackgroundTaskManager *)sharedInstance
...
+}
+
+- (void)stopOperationWithType:(MSIDBackgroundTaskType)type caller:(nonnull id)caller
+{
+    MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
+    
+    if ((backgroundTaskData.backgroundTaskId == UIBackgroundTaskInvalid) || (![caller isEqual:backgroundTaskData.callerReference]))
+    {
+        // Background task not started or background task can be stopped only if it started it
+        return;
```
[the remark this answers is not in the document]
ameyapat: Sure, I think that would be helpful but I think modifying the log that started and stopped operation successfully would be just as useful. I'll modify that too.

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
ameyapat: You're right, I had a different implementation in mind earlier and missed changing this back. isEqual will compare hash value instead of reference.

ameyapat: Earlier implementation had it set as invalid id instead of removing from cache. I too think it makes sense to remove it from cache when stopOperation is called.

ameyapat:
Marking as WIP, there is still a case where this logic can cause a broker request to lose background thread protection. Will update PR with resolution and remove WIP

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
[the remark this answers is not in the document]
ameyapat: Added it as a separate file

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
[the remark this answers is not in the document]
ameyapat: Removed the need to pass in a reference.

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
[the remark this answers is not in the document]
ameyapat: Added log in stopOperationWithType: to print when operation is being stopped.

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
ameyapat: Removed need to pass reference

ameyapat: Modified code to remove entry from cache instead of marking it as Invalid.

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
```diff
@@ -65,55 +66,62 @@ + (MSIDBackgroundTaskManager *)sharedInstance
...
 /*
  Background task execution:
- https://forums.developer.apple.com/message/253232#253232
+ https://developer.apple.com/forums/thread/85066
  */
 
 - (void)startOperationWithType:(MSIDBackgroundTaskType)type
 {
-    UIBackgroundTaskIdentifier backgroundTask = [self backgroundTaskWithType:type];
+    MSIDBackgroundTaskData *backgroundTaskData = [self backgroundTaskWithType:type];
```
jasoncoolmax, line 74: Seems the implementation of this class is not thread-safe?

E.g., if there are two silent calls going on at the same time. The execution sequence could be:
1. Silent call A execute line 103 of this file;
2. Silent call B execute line 74;
3. Silent call A execute line 104-105;
4. Silent call B execute line 76-80.

In the above sequence, Silent call B loses background task protection. What do you think?

ameyapat: In this example, 
1. Silent call A calls method to stop bg task operation.
2. Silent call B gets reference to Bg task that was started by A from cache at line 74.
3. Silent call A ends bg task and removes bg task data object from cache. But in 2. since the bg task ref is still retained, it won't be deallocated.
4. Silent call B will not start a new bg task.

Great catch @jasoncoolmax ! I think in step 2. when silent call B gets reference to bg task data, if I mark it as a weak reference, that might resolve this. What do you think?

jasoncoolmax: I am fine with the weak reference solution, given that using @ synchronized may sacrifice some performance here.

jasoncoolmax (approved):

IdentityCore/src/webview/background/ios/MSIDBackgroundTaskManager.m
ameyapat: @jasoncoolmax Added test case to test if network requests might loose background task protection in the event that multiple nw requests simultaneously call for bg task protection through threads. I think OneDrive/Office does this to refresh tokens in its cache periodically via multiple threads.

The test case was failing without synchronized block for start & stop bg task operation methods. I've added synchronized blocks per your suggestion to both methods so that no 2 threads enter those methods simultaneously.
Please let me know if I missed something :)

jasoncoolmax (approved):

PR #1123, opened 2022-06-30
[WIP] Force legacy broker fallback for ZTP + add "userless registration needed" as a request param

mipetriu:
## Proposed changes

In the Zero-Touch Provisioning public preview flow, users will expect to sign in from any shared device-participating app, and have the SSO extension automatically handle shared device registration. In order to do that, we need to force an automatic UI flip to the Authenticator app, which will automatically trigger userless registration when the Authenticator app is opened. A good way to achieve this is to force a fallback to the legacy broker flow when userless registration is needed - the SSO extension will detect if userless registration is needed, and return an "expected" error to MSAL. MSAL will recognize this error as a trigger for the userless registration flow, and trigger a fallback to the legacy broker flow.

Later on, the Authenticator app will be notified in the openURL request that userless registration is needed through the use of a new extra query param: "userless_reg". This will trigger any necessary UI actions on the Authenticator app side.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/controllers/broker/MSIDSSOExtensionInteractiveTokenRequestController.m
+++ b/IdentityCore/src/controllers/broker/MSIDSSOExtensionInteractiveTokenRequestController.m
@@ -32,6 +32,8 @@
 
 @implementation MSIDSSOExtensionInteractiveTokenRequestController
 
+NSString* const MSIDExpectedErrorDomain = @"ExpectedErrorDomain";
+
 - (instancetype)initWithInteractiveRequestParameters:(MSIDInteractiveTokenRequestParameters *)parameters
                                 tokenRequestProvider:(id<MSIDTokenRequestProviding>)tokenRequestProvider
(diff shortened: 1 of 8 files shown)


IdentityCore/src/controllers/broker/MSIDSSOExtensionInteractiveTokenRequestController.m
```diff
@@ -32,6 +32,8 @@
 
 @implementation MSIDSSOExtensionInteractiveTokenRequestController
 
+NSString* const MSIDExpectedErrorDomain = @"ExpectedErrorDomain";
```
antrix1989, line 35: Why do we need a separate domain for this error? This flow seems to be similar to MSIDErrorServerProtectionPoliciesRequired error

PR #1127, opened 2022-07-29
Make a property public for MSAL CPP to overwrite -(void)presentView

fengga:
## Proposed changes

MSAL CPP need to implement form sheet UI style on iPad. The only way is to overwrite -(void)presentView in MSIDWebviewUIController.m, but -(void)presentView is using a private property propertypresentInParentController.

Since MSIDOAuth2EmbeddedWebviewController which extends MSIDWebviewUIController is also used by MSAL CPP, adding a MSIDWebviewUIController+private will be not very applicable.

The easiest way is to make propertypresentInParentController a public property.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/webview/embeddedWebview/ui/MSIDWebviewUIController.h
+++ b/IdentityCore/src/webview/embeddedWebview/ui/MSIDWebviewUIController.h
@@ -46,7 +46,7 @@ NSWindowController
 #if TARGET_OS_IPHONE
 @property (nonatomic, weak) UIViewController *parentController;
 @property (nonatomic) UIModalPresentationStyle presentationType;
-@property (nonatomic) BOOL presentInParentController;
+@property (nonatomic, readonly) BOOL presentInParentController;
 #endif
 
 + (WKWebViewConfiguration *)defaultWKWebviewConfiguration;
--- a/IdentityCore/src/webview/embeddedWebview/ui/MSIDWebviewUIController.h
+++ b/IdentityCore/src/webview/embeddedWebview/ui/MSIDWebviewUIController.h
@@ -46,6 +46,7 @@ NSWindowController
 #if TARGET_OS_IPHONE
 @property (nonatomic, weak) UIViewController *parentController;
 @property (nonatomic) UIModalPresentationStyle presentationType;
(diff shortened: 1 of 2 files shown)


IdentityCore/src/webview/embeddedWebview/ui/MSIDWebviewUIController.h
```diff
@@ -46,6 +46,7 @@ NSWindowController
 #if TARGET_OS_IPHONE
 @property (nonatomic, weak) UIViewController *parentController;
 @property (nonatomic) UIModalPresentationStyle presentationType;
+@property (nonatomic) BOOL presentInParentController;
```
oldalton, line 49: You also need to update macOS implementation for the same

oldalton: Also, I'm a bit confused. Why don't you just pass in webview?

fengga: I just double checked, the macOS version of MSIDWebviewUIController doesn't have presentInParentController.

For your second question - I need to overwrite the -(void)presentView method in MSAIWebviewUIController and MSAIOAuth2EmbeddedWebviewController in MSAL CPP, so I need access to presentInParentController in MSAL CPP.

original:
b2811861c1c412cd649245f20055911

new:
cf118b6fc3b29ee82dfb4b7a2ecc5cc

Office is doing this manually every time after they consume OneAuth/MSAL, and we have an email thread that OneAuth/MSAL should do this for all partners.

oldalton: Ok, then you need to make this property readonly in the public header file + mark as iOS only. You can keep it as a readwrite in the .m file.

fengga: @oldalton, thanks!

kaisong1990 (approved):

PR #1259, opened 2023-06-30
Added code that removes expired AT for apple storage.

kaisong1990:
## Proposed changes

@dorinung:

AB#2613823

Added code that removes expired AT for apple storage.
(removeExpiredCredentialsWithQuerys). This method is called from deleteAccessTokens which in turn is called from writeCredentials. Note that is only called from MSAIStorageManagerImpl.m so it will not be part of any other IdentityCore code paths. It relies on the addition getCredentialsWithQuery and matchesWithRealm relying on new value MSIDAny in MSAIMSIDComparisonOptions to return all entries for homeAccountId, environment, realm and clientId in cache query. The returned entries are then tested for expiration. Expired tokens are deleted using removeAllCredentials call.

Here is an MSAL draft PR that shows end to end changes with both XPLAT and apple storage changes.

https://github.com/AzureAD/microsoft-authentication-library-for-cpp/pull/3612

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
+++ b/IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
@@ -313,6 +313,51 @@
                                 error:error];
 }
 
+// Remove credentials
+- (BOOL)removeExpiredAccessTokensCredentialsWithQuery:(nonnull MSIDDefaultCredentialCacheQuery *)cacheQuery
+                          context:(nullable id<MSIDRequestContext>)context
+                            error:(NSError * _Nullable * _Nullable)error
+{
+    assert(cacheQuery);
+    cacheQuery.targetMatchingOptions = MSIDAny;
+    cacheQuery.matchAnyCredentialType = NO;
+    NSDate *currentDataAndTime = [NSDate date];
+    NSMutableArray *resultsToDelete = [NSMutableArray array];
+    NSString *className = NSStringFromClass(self.class);
+
+    MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, context, @"(Default cache) Removing expired credentials with type %@, environment %@, realm %@, clientID %@, unique user ID %@", [MSIDCredentialTypeHelpers credentialTypeAsString:cacheQuery.credentialType], cacheQuery.environment, cacheQuery.realm, cacheQuery.clientId, MSID_PII_LOG_TRACKABLE(cacheQuery.homeAccountId));
+
+    if (cacheQuery.exactMatch)
+    {
+        return NO;
+    }
+    
+    if (cacheQuery.credentialType != MSIDAccessTokenType)
+    {
+        return NO;
+    }
+
+    NSArray<MSIDCredentialCacheItem *> *matchedCredentials = [self getCredentialsWithQuery:cacheQuery context:context error:error];
+    
+    if (!matchedCredentials) return NO;
+    
+    // Check for expiry here to leave only expired ones in matchedCredentials
+    for (MSIDCredentialCacheItem *cacheItem in matchedCredentials)
+    {
+        NSComparisonResult result = [currentDataAndTime compare:[cacheItem expiresOn]];
+        if (result == NSOrderedDescending)
+        {
+            [resultsToDelete addObject:cacheItem];
+        }
+    }
+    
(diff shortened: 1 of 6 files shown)


juan-arias (approved):

IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
```diff
@@ -313,6 +313,51 @@ - (BOOL)removeCredentialsWithQuery:(nonnull MSIDDefaultCredentialCacheQuery *)ca
...
+    }
+    
+    if (cacheQuery.credentialType != MSIDAccessTokenType)
+    {
+        return NO;
+    }
+
+    NSArray<MSIDCredentialCacheItem *> *matchedCredentials = [self getCredentialsWithQuery:cacheQuery context:context error:error];
+    
+    if (!matchedCredentials) return NO;
```
ameyapat, line 342: Change to if (!matchedCredentials.count) return NO; to check for empty array

IdentityCore/tests/MSIDAccountCredentialsCacheTests.m
```diff
@@ -2755,6 +2755,41 @@ - (void)testRemoveCredentialsWithQuery_whenQueryIsNotExactMatch_andAccessTokensQ
...
+
+    MSIDCredentialCacheItem *token2 = [self createTestAccessTokenCacheItem];
+    token2.homeAccountId = @"uid.utid1";
+    token2.realm = @"contoso2.com";
+    token2.expiresOn = [[NSDate date] dateByAddingTimeInterval:60];
+    [self saveItem:token2];
+
+    [self saveItem:[self createTestRefreshTokenCacheItem]];
+    
+    [NSThread sleepForTimeInterval:5.0];
```
ameyapat, line 2773: Instead of sleeping for 5 seconds can the token.expiresOn be set 5 seconds earlier from now : [NSDate dateWithTimeIntervalSinceNow:-5.0]?

ameyapat (approved):
Approved with suggested changes

IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
```diff
@@ -313,6 +313,51 @@ - (BOOL)removeCredentialsWithQuery:(nonnull MSIDDefaultCredentialCacheQuery *)ca
...
+    assert(cacheQuery);
+    cacheQuery.targetMatchingOptions = MSIDAny;
+    cacheQuery.matchAnyCredentialType = NO;
+    NSDate *currentDataAndTime = [NSDate date];
+    NSMutableArray *resultsToDelete = [NSMutableArray array];
+    NSString *className = NSStringFromClass(self.class);
+
+    MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, context, @"(Default cache) Removing expired credentials with type %@, environment %@, realm %@, clientID %@, unique user ID %@", [MSIDCredentialTypeHelpers credentialTypeAsString:cacheQuery.credentialType], cacheQuery.environment, cacheQuery.realm, cacheQuery.clientId, MSID_PII_LOG_TRACKABLE(cacheQuery.homeAccountId));
+
+    if (cacheQuery.exactMatch)
```
Veena11, line 330: Why do we return NO when query is exactMatch ?

kaisong1990: @dorinung can you provide some context? My understanding is that when the query requires exact match, we don't perform removal during look up.

IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
dorinung: The same pattern is used removeCredentialsWithQuery above. It is a null check. getCredentialsWithQuery returns nil on errors.

IdentityCore/src/cache/accessor/MSIDAccountCredentialCache.m
dorinung: @kaisong1990  That is correct. We only want this code to run for non exact matches.

PR #1365, opened 2024-04-30
Add visionOS support to IdentityCore iOS target

mipetriu:
## Proposed changes

This PR adds visionOS support to the IdentityCore iOS target, so that IdentityCore iOS compiles on iOS, iPadOS, and visionOS. It also adds code changes needed to compile on visionOS. The biggest change is that visionOS no longer supports SafariViewController, so we need to ensure that all references to it are behind OS macros in order for the target to build.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/webview/systemWebview/session/MSIDSystemWebviewController.m
```diff
@@ -237,10 +243,11 @@ - (void)dismiss
         return safariController;
     }
 #else
-    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Couldn't create session on macOS. Safari allowed flag %d", safariAllowed);
+    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, nil, @"Couldn't create session on macOS or visionOS. Safari allowed flag %d", safariAllowed);
```
Veena11, line 246: if target OS is Vision, we are returning nil at line 230, so is this log change correct ?

IdentityCore/src/webview/embeddedWebview/challangeHandlers/ios/MSIDCertAuthHandler.m
```diff
@@ -92,7 +92,9 @@ + (BOOL)completeCertAuthChallenge:(NSURL *)endUrl
 
 + (void)resetHandler
 {
+#if TARGET_OS_IPHONE && !MSID_EXCLUDE_SYSTEMWV
```
Veena11, line 95: why we added this extra check for system webview here ?

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -81,11 +81,15 @@ @implementation MSIDWebviewFactory
...
                 break;
         }
-    
+
+#if !MSID_EXCLUDE_SYSTEMWV
     return [self systemWebviewFromConfiguration:configuration
                        useAuthenticationSession:useSession
                       allowSafariViewController:allowSafariViewController
                                         context:context];
+#else
```
Veena11, line 90: are we returning nil only if it is system webview  for all targets ?

Veena11 (commented):
Added some questions for my learning

IdentityCore/src/oauth2/MSIDWebviewFactory.m
mipetriu: Yes, any partner regardless of platform may choose to avoid system webview entirely.  These `!MSID_EXCLUDE_SYSTEMWV` changes are added in a few places where setting it was not working. Other changes that add `#if defined TARGET_OS_VISION && TARGET_OS_VISION` are meant to specifically remove SafariViewController because it is completely removed on visionOS

In this specific case the `[systemWebviewFromConfiguration]` method itself is wrapped in `#if !MSID_EXCLUDE_SYSTEMWV`, but the call to it here was not. It's safer to wrap the call itself as well so we don't call a non-existent method

IdentityCore/src/webview/embeddedWebview/challangeHandlers/ios/MSIDCertAuthHandler.m
mipetriu: This is a suggested change from Teams based on testing with the system webview disabled. I can see that all other instances of `s_certAuthInProgress` being updated are wrapped in this macro

IdentityCore/src/webview/systemWebview/session/MSIDSystemWebviewController.m
mipetriu: That's true, I added the visionOS logic above later. I'll add a log line up there too

PR #1412, opened 2024-08-19
IdentityCore visionOS support

mipetriu:
## Proposed changes

This PR adds visionOS support to the IdentityCore iOS target, so that IdentityCore iOS compiles on iOS, iPadOS, and visionOS. It also adds code changes needed to compile on visionOS. The biggest change is that visionOS no longer supports SafariViewController, so we need to ensure that all references to it are behind OS macros in order for the target to build.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/util/ios/MSIDAppExtensionUtil.m
```diff
@@ -77,7 +77,11 @@ + (void)sharedApplicationOpenURL:(NSURL*)url
 #pragma clang diagnostic ignored "-Warc-performSelector-leaks"
     
     [MSIDMainThreadUtil executeOnMainThreadIfNeeded:^{
+#if defined TARGET_OS_VISION && TARGET_OS_VISION
```
Veena11, line 80: nit: this check could be defined as a new macro something like MSID_TARGET_OS_VISION, so you could avoid doing this && operation everywhere

IdentityCore/src/webview/systemWebview/session/MSIDSystemWebViewControllerFactory.m
```diff
@@ -100,10 +100,18 @@ + (MSIDWebviewType)availableWebViewTypeWithPreferredType:(MSIDWebviewType)prefer
...
+#endif
+
+#if !MSID_EXCLUDE_SYSTEMWV
     return [[MSIDSafariViewController alloc] initWithURL:startURL
                                         parentController:parentController
                                         presentationType:presentationType
                                                  context:context];
+#else
+    return nil;
+#endif
```
Veena11, line 114: you could check #if !MSID_EXCLUDE_SYSTEMWV && !(defined TARGET_OS_VISION && TARGET_OS_VISION) to return SafariViewController and return nil otherwise

Veena11 (approved):

IdentityCore/src/webview/systemWebview/session/MSIDSystemWebViewControllerFactory.m
mipetriu: Makes sense, good suggestion. Updated

IdentityCore/src/util/ios/MSIDAppExtensionUtil.m
jasoncoolmax: for my own learning, why do we need `#if defined TARGET_OS_VISION && TARGET_OS_VISION` but not `#if TARGET_OS_VISION`? :)

IdentityCore/src/webview/embeddedWebview/challangeHandlers/ios/MSIDCertAuthHandler.m
```diff
@@ -97,7 +97,9 @@ + (BOOL)isCertAuthInProgress
 
 + (void)resetHandler
 {
+#if TARGET_OS_IPHONE && !MSID_EXCLUDE_SYSTEMWV
     s_certAuthInProgress = NO;
+#endif
```
jasoncoolmax, line 102: Do we need `#if !MSID_EXCLUDE_SYSTEMWV` for line 83 and 93? Seems `s_certAuthInProgress` is only defined when !MSID_EXCLUDE_SYSTEMWV

IdentityCore/src/util/ios/MSIDAppExtensionUtil.m
mipetriu: I previously tested these changes on an older version of XCode before the visionOS simulator was introduced, and I saw build errors due to using an undefined macro. The "defined" check ensures that we can still build on XCode <15.2

IdentityCore/src/webview/embeddedWebview/challangeHandlers/ios/MSIDCertAuthHandler.m
mipetriu: Yes, this is needed because the declaration of the variable itself is wrapped in the macro as well, so there is actually a build error when MSID_EXCLUDE_SYSTEMWV is set without this

mipetriu: And regarding this:

> for my own learning, why do we need #if defined TARGET_OS_VISION && TARGET_OS_VISION but not #if TARGET_OS_VISION? :)

This is for backwards compatibility with older XCode versions that may not have visionOS support

jasoncoolmax (approved):

PR #1434, opened 2024-10-16
Assign completion block before perform request

xinyuxu1026:
## Proposed changes

Assign completion block before perform request and fix potential threading issue.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/requests/broker/MSIDSSOExtensionGetAccountsRequest.m
+++ b/IdentityCore/src/requests/broker/MSIDSSOExtensionGetAccountsRequest.m
@@ -148,9 +148,9 @@
     self.authorizationController.delegate = self.extensionDelegate;
     
     self.requestSentDate = [NSDate date];
-    [self.authorizationController msidPerformRequests];
-    
+
     self.requestCompletionBlock = completionBlock;
+    [self.authorizationController msidPerformRequests];
 }
 
 #pragma mark - AuthenticationServices
--- a/IdentityCore/src/requests/broker/MSIDSSOExtensionGetDeviceInfoRequest.m
+++ b/IdentityCore/src/requests/broker/MSIDSSOExtensionGetDeviceInfoRequest.m
@@ -131,9 +131,9 @@
     self.authorizationController = [self controllerWithRequest:ssoRequest];
     self.authorizationController.delegate = self.extensionDelegate;
     self.requestSentDate = [NSDate date];
-    [self.authorizationController msidPerformRequests];
-    
+
     self.requestCompletionBlock = completionBlock;
+    [self.authorizationController msidPerformRequests];
 }
 
 #pragma mark - AuthenticationServices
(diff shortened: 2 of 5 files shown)


hieunguyenmsft (approved):
LGTM, please update changelog

xinyuxu1026:
@oldalton , here is the test i did about this code:

```
[self.authorizationController msaimsidPerformRequests];
// wait 10s
self.requestCompletionBlock = completionBlock;
```
i see thread hang and no callback
 
```
self.requestCompletionBlock = completionBlock;
// wait 10s
[self.authorizationController msaimsidPerformRequests];
```
i got the callback after 10s. 
so there might be hang issue in the original code.

oldalton:
> @oldalton , here is the test i did about this code:
> 
> ```
> [self.authorizationController msaimsidPerformRequests];
> // wait 10s
> self.requestCompletionBlock = completionBlock;
> ```
> 
> i see thread hang and no callback
> 
> ```
> self.requestCompletionBlock = completionBlock;
> // wait 10s
> [self.authorizationController msaimsidPerformRequests];
> ```
> 
> i got the callback after 10s. so there might be hang issue in the original code.

Thank you. Can you tell me more about your test? Is there anything else blocking the original thread?

xinyuxu1026:
@oldalton , I tested using the silent token request, the main thread is not blocking but OneAuth is not getting the callback.

oldalton:
> @oldalton , I tested using the silent token request, the main thread is not blocking but OneAuth didn't get the callback.

Ok, I'm fine to merge this change @xinyuxu1026 to unblock you, but I'd still like you to understand why this is happening for you. Same isn't happening for MSAL Obj-C, so there's clearly some difference in the calling patterns. I think we need to be aware of those differences to avoid any similar pitfalls in the future. Could you create a follow-up task to investigate and be able to explain why it's happening?

oldalton (approved):

kaisong1990:
> > @oldalton , I tested using the silent token request, the main thread is not blocking but OneAuth didn't get the callback.
> 
> Ok, I'm fine to merge this change @xinyuxu1026 to unblock you, but I'd still like you to understand why this is happening for you. Same isn't happening for MSAL Obj-C, so there's clearly some difference in the calling patterns. I think we need to be aware of those differences to avoid any similar pitfalls in the future. Could you create a follow-up task to investigate and be able to explain why it's happening? You can reach out to @kaisong1990 in my team for additional help once he's back in the office.

Hi @xinyuxu1026 @githubursul there is no harm in making that change. However, by adding the 10s delay before assigning the completion block, there seems to be expected that the OenAuth did not get callback.

For example, broker handles a silent request, and the response is sent back to client/OneAuth within 200 ms, due to the delay, the completion handler is not assigned yet so the final response will not be passed back to OneAuth layer.

Instead, I would suggest adding a breakpoint at `MSIDSSOExtensionTokenRequestDelegate.m` L35 and compare if client/OneAuth received response or not in original and suggested changes with 10 secs delay in below method.

`- (void)authorizationController:(__unused ASAuthorizationController *)controller didCompleteWithAuthorization:(ASAuthorization *)authorization
`

xinyuxu1026:
@oldalton , I had a debug session with @kaisong1990 , and we found that in my test, in silent token request, main thread is not blocked, but there is a thread waiting even after 10s. 
Screenshot 2024-10-21 at 5 18 28 PM

With my change, there is no thread hang after waiting 10s. There might be small chance that the broker completes the request so fast that the completion block is not assigned yet.

PR #1444, opened 2024-11-15
Add pii around upn for wp registration web response

kaisong1990:
## Proposed changes

Add PII around upn

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
@@ -122,7 +122,7 @@
             MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters, @"Workplace join Upgrade registration is required.");
             
             NSMutableDictionary *additionalInfo = [NSMutableDictionary new];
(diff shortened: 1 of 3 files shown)


oldalton (changes requested):
We shouldn't use _PII_NULLIFY for actual code, since we rely on user info as part of the flow. We should only do _PII_NULLIFY in logs

kaisong1990:
Hi @oldalton do we need to check into this release or if we can wait till the next one?

kaisong1990:
> We shouldn't use _PII_NULLIFY for actual code, since we rely on user info as part of the flow. We should only do _PII_NULLIFY in logs

Hi @oldalton I' benchmarked the performance and the difference is below 1 ms. The NSRange solution is slightly better and it is currently used in `MSIDMaskedUsernameLogParameter` so I switched with the similar logic

swasti29:
@kaisong1990  Please close the PR if not valid anymore.

PR #1489, opened 2025-03-05
Add feature flags provider to be controlled from broker

juan-arias:
## Proposed changes

Add feature flags provider to be controlled from broker

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/MSIDBrokerFlightProvider.m
```diff
@@ -0,0 +1,110 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.  
+
+#import "MSIDBrokerFlightProvider.h"
+#import "NSJSONSerialization+MSIDExtensions.h"
+
+@interface MSIDBrokerFlightProvider()
+
+@property (nonatomic, nullable) NSDictionary *clientFlightsPayload;
```
kaisong1990, line 30: Question: is it necessary to make this thread safe? Will the same instance be accessed through different threads?

IdentityCore/src/MSIDBrokerFlightProvider.h
```diff
@@ -0,0 +1,40 @@
...
+// THE SOFTWARE.  
+
+#import <Foundation/Foundation.h>
+#import "MSIDFlightManager.h"
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSIDBrokerFlightProvider : NSObject <MSIDFlightManagerInterface>
+
+- (instancetype)initWithBase64EncodedFlightsPayload:(nullable NSString *)base64EncodedFlightsPayload;
```
kaisong1990, line 33: The init method is in NS_ASSUME_NONNULL_BEGIN block, but implementation can return nil. Please consider adding `nullable`

IdentityCore/src/MSIDBrokerFlightProvider.m
```diff
@@ -0,0 +1,110 @@
...
+
+- (nullable NSArray<NSString *> *)stringsForFlightKey:(nonnull NSString *)flightKey
+{
+    if (self.clientFlightsPayload)
+    {
+        id value = self.clientFlightsPayload[flightKey];
+        
+        if ([value isKindOfClass:[NSArray class]])
+        {
+            return value;
```
kaisong1990, line 102: nit: what if the elemet in the array is not `NSString`

IdentityCore/src/requests/MSIDTokenResponseHandler.m
```diff
@@ -89,6 +92,16 @@ - (void)handleTokenResponse:(MSIDTokenResponse *)tokenResponse
         return;
     }
     
+#if !AD_BROKER
```
kaisong1990, line 95: nit: This does not sound like part of the token handler, is it possible to move the logic into `MSIDDeviceInfo`

IdentityCore/src/requests/MSIDTokenResponseHandler.m
```diff
@@ -89,6 +92,16 @@ - (void)handleTokenResponse:(MSIDTokenResponse *)tokenResponse
...
+    // Save client flights if available
+    if (tokenResponse.clientFlights)
+    {
+        MSIDBrokerFlightProvider *flightProvider = [[MSIDBrokerFlightProvider alloc] initWithBase64EncodedFlightsPayload:tokenResponse.clientFlights];
+        
+        [MSIDFlightManager sharedInstance].flightProvider = flightProvider;
```
kaisong1990, line 101: The `flightProvider` in `MSIDFlightManager` may not be thread safe. Since this is a global instance. While one thread is reading the `flightProvider`, and another SsoExt callback is assigning value here, will cause crash.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
```diff
@@ -154,6 +157,16 @@ - (MSIDTokenResult *)handleBrokerResponseWithURL:(NSURL *)response sourceApplica
         return nil;
     }
     
+#if !AD_BROKER
+    // Save client flights if available
+    if (brokerResponse.clientFlights)
+    {
+        MSIDBrokerFlightProvider *flightProvider = [[MSIDBrokerFlightProvider alloc] initWithBase64EncodedFlightsPayload:brokerResponse.clientFlights];
+        
+        [MSIDFlightManager sharedInstance].flightProvider = flightProvider;
```
kaisong1990, line 166: ditto.

Also if moving the logic into a lower leverl, may not need to duplicate

kaisong1990 (commented):
Hi @juan-arias I have put some comments

IdentityCore/src/MSIDBrokerFlightProvider.m
juan-arias: I think it is safe, beacuse it will be written once during initialization; but then can be read through multiple threads, but that's ok.

IdentityCore/src/MSIDBrokerFlightProvider.h
juan-arias: Updated

IdentityCore/src/MSIDBrokerFlightProvider.m
juan-arias: It was returning `nil`. I've now updated this method to return a single NSString value from the flight; in case the value from ECS is not a string, this method is also returning `nil`.

IdentityCore/src/requests/MSIDTokenResponseHandler.m
juan-arias: Updated

IdentityCore/src/requests/MSIDTokenResponseHandler.m
juan-arias: Thank you Kai, updated.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
juan-arias: Updated

IdentityCore/src/MSIDFlightManager.m
```diff
@@ -25,33 +25,75 @@
...
+    {
+        _synchronizationQueue = [self initializeDispatchQueue];
+    }
+    return self;
+}
+
+- (dispatch_queue_t)initializeDispatchQueue
+{
+    NSString *queueName = [NSString stringWithFormat:@"com.microsoft.msidflightmanager-%@", [NSUUID UUID].UUIDString];
+    return dispatch_queue_create([queueName cStringUsingEncoding:NSASCIIStringEncoding], DISPATCH_QUEUE_SERIAL);
```
kaisong1990, line 61: nit: please consider to use concurrent instead of DISPATCH_QUEUE_SERIAL

IdentityCore/src/MSIDFlightManager.m
```diff
@@ -25,33 +25,75 @@
...
+        self->_flightProvider = flightProvider;
+    });
+}
+
 #pragma mark - MSIDFlightManagerInterface
 
 - (BOOL)boolForKey:(nonnull NSString *)flightKey 
 {
-    if (self.flightProvider) { return [self.flightProvider boolForKey:flightKey]; }
+    __block BOOL result = NO;
```
kaisong1990, line 75: If using concurrent queue, read can happen in parallel instead of waiting for each other. You can still keep  dispatch_barrier_async to make sure write only happened when the queue is not executing other tasks

kaisong1990 (approved):
LGTM, a minor suggestions may help to improve the performance

antrix1989 (approved):

PR #1516, opened 2025-04-29
Update Xpc silent controler error handling as fallback

kaisong1990:
## Proposed changes

In this PR, a new protocol is added to the base MSIDRequestControlling.h. This new protocol is used to decide whether the current fallback request controller should continue acquiring token based on the passed in error from the main controller.

1. Updated XPC silent controller when work as a SsoExt fallback controller, only below 4 error code will trigger a fallback:
        case ASAuthorizationErrorNotHandled:
        case ASAuthorizationErrorUnknown:
        case ASAuthorizationErrorFailed:
        case MSIDErrorSSOExtensionUnexpectedError:
2. Update the original error handling logic that either returns the main token request controller error or the valid token result from the fallback controller. This also aligns with OA

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/controllers/MSIDSilentController.m
```diff
@@ -143,19 +143,26 @@ - (void)acquireTokenWithRequest:(MSIDSilentTokenRequest *)request
...
+            // (!result && !error) is a special case when using local silent controller and to skip local refreh token (broker first flow). In this case, the main controller is the 1st fallback controller, and should return the error
+            if (fallResult || (!result && !error))
             {
-                completionBlock(result, error);
+                completionBlock(fallResult, fallError);
             }
             else
             {
-                completionBlock(fallResult, fallError);
+                completionBlock(result, error);
```
antrix1989, line 154: why do we call fallback if we already have result?

IdentityCore/src/controllers/MSIDRequestControlling.h
```diff
@@ -31,6 +31,7 @@ NS_ASSUME_NONNULL_BEGIN
 @protocol MSIDRequestControlling <NSObject>
 
 - (void)acquireToken:(nonnull MSIDRequestCompletionBlock)completionBlock;
+- (void)shouldSkipAcquireTokenBasedOn:(nonnull NSError *)error;
```
antrix1989, line 34: 1. can we add this flag directly  to "acquireToken" method?
2. Nit: Can we rename it to "skipAcquire..." instead of "should..."?"should" sounds like a dynamic condition to me, while here we know for sure that we want to skip.

IdentityCore/src/controllers/MSIDSilentController.m
kaisong1990: In this case, the result is nil, the valid part is error only. I can send `nil` specifically

IdentityCore/src/controllers/MSIDRequestControlling.h
ameyapat: +1, methods with 'should' prefix usually indicate return type as BOOL but this does not seem to be the case here.

IdentityCore/src/controllers/broker/mac/MSIDXpcSilentTokenRequestController.m
```diff
@@ -55,4 +79,24 @@ + (BOOL)canPerformRequest
     }
 }
 
+- (void)shouldSkipAcquireTokenBasedOn:(NSError *)error
+{
+    self.shouldSkipAcquireToken = YES;
```
ameyapat, line 84: What if error is nil? This method will log incorrect info which might confuse in case of issue debugging

IdentityCore/src/controllers/broker/mac/MSIDXpcSilentTokenRequestController.m
```diff
@@ -55,4 +79,24 @@ + (BOOL)canPerformRequest
...
+    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters, @"Looking if we should fallback to Xpc controller, error: %ld error domain: %@.", (long)error.code, error.domain);
+    // If it is MSIDErrorDomain and Sso Extension returns unexpected error, we should fall back to local controler and unblock user
+    if (![error.domain isEqualToString:ASAuthorizationErrorDomain] && ![error.domain isEqualToString:MSIDErrorDomain]) {
+    }
```
ameyapat, line 88: What is this supposed to do? Did you miss something?

IdentityCore/src/controllers/broker/mac/MSIDXpcSilentTokenRequestController.m
```diff
@@ -55,4 +79,24 @@ + (BOOL)canPerformRequest
...
+    
+    switch (error.code)
+    {
+        case ASAuthorizationErrorNotHandled:
+        case ASAuthorizationErrorUnknown:
+        case ASAuthorizationErrorFailed:
+        case MSIDErrorSSOExtensionUnexpectedError:
+            self.shouldSkipAcquireToken = NO;
```
ameyapat, line 96: Can the controller be reused for multiple requests? If it is, controller's shouldSkipAcquireToken might get set to NO in 1st request and remain NO for second request

IdentityCore/src/controllers/broker/mac/MSIDXpcSilentTokenRequestController.m
```diff
@@ -28,20 +28,44 @@
 #import "MSIDLogger+Internal.h"
 #import "MSIDXpcProviderCache.h"
 
+@interface MSIDXpcSilentTokenRequestController ()
+
+// shouldSkipXpcRequest is used when the fallback is not needed for non ASAuthorizationErrorDomain or MSIDErrorSSOExtensionUnexpectedError
+@property (nonatomic) BOOL shouldSkipAcquireToken;
```
ameyapat, line 34: The comment says shouldSkipXpcRequest but property is named shouldSkipAcquireToken. Can you rename property to shouldSkipCallingXpc? shouldSkipAcquireToken is confusing

IdentityCore/src/controllers/broker/mac/MSIDXpcSilentTokenRequestController.m
```diff
@@ -28,20 +28,44 @@
...
+            completionBlock(result, error);
+        };
+        
+        __auto_type request = [self.tokenRequestProvider silentXpcTokenRequestWithParameters:self.requestParameters
+                                                                                forceRefresh:self.forceRefresh];
+        [self acquireTokenWithRequest:request completionBlock:completionBlockWrapper];
+    }
+    else
+    {
+        // self.fallbackController cannot be nil here as it has been validated from caller
```
ameyapat, line 57: Is this comment inaccurate? If we know fallbackController cannot be nil here, why do a nil check and log in the code? Please remove the comment as it isn't accurate according to the code following it

swasti29:
@kaisong1990 Feel free to reopen if it is still relevant.

PR #1520, opened 2025-05-01
 Handle account removed suberror in brokers and MSALs

antonioalwan:
## Proposed changes

Ensuring that any remaining artifacts on the device for deleted user accounts are removed when ESTS returns the suberror=user_account_deleted. This fix should be applied in the common core where Brokers and MSALs are addressed.
Added TestFlight entry: "disable_rm_metadata"
## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

WILink: https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3168316

### DIFF
--- a/IdentityCore/src/MSIDConstants.h
+++ b/IdentityCore/src/MSIDConstants.h
@@ -215,4 +215,15 @@ extern NSString * _Nonnull const MSID_FLIGHT_USE_V2_WEB_RESPONSE_FACTORY;
 extern NSString * _Nonnull const MSID_FLIGHT_SUPPORT_DUNA_CBA;
 extern NSString * _Nonnull const MSID_FLIGHT_CLIENT_SFRT_STATUS;
 
+/**
+ * Flight to indicate if remove account artifacts should be disabled
+ * Owner: Antonio
+ * Link: N/A - No ECS flag created as this is a disable flight to be created on demand
+ * Created Date: N/A
+ * Status: Not started
+ * Fully Allocated: Not Started
+ * WorkItem: 3168316
+ */
+extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_REMOVE_ACCOUNT_ARTIFACTS;
+
 #define METHODANDLINE   [NSString stringWithFormat:@"%s [Line %d]", __PRETTY_FUNCTION__, __LINE__]
--- a/IdentityCore/src/requests/MSIDSilentTokenRequest.m
+++ b/IdentityCore/src/requests/MSIDSilentTokenRequest.m
@@ -699,10 +699,12 @@ typedef NS_ENUM(NSInteger, MSIDRefreshTokenTypes)
     return NO;
 }
 
-- (BOOL)shouldRemoveAccountArtifacts:(__unused NSError *)serverError
+- (BOOL)shouldRemoveAccountArtifacts:(nonnull NSError *)serverError
 {
-    NSAssert(NO, @"Abstract method. Should be implemented in a subclass");
-    return NO;
+    // Removing account artifacts on invalid_grant + user_deleted_account suberror combination
+    MSIDErrorCode oauthError = MSIDErrorCodeForOAuthError(serverError.msidOauthError, MSIDErrorInternal);
+    NSString *subError = serverError.msidSubError;
+    return oauthError == MSIDErrorServerInvalidGrant && [subError isEqualToString:MSIDServerErrorUserAccountDeleted];
 }
 
 - (id<MSIDCacheAccessor>)tokenCache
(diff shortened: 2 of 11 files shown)


IdentityCore/src/requests/sdk/adal/MSIDLegacySilentTokenRequest.m
```diff
@@ -100,6 +100,14 @@ - (BOOL)shouldRemoveRefreshToken:(NSError *)serverError
     return oauthError == MSIDErrorServerInvalidGrant;
 }
 
+- (BOOL)shouldRemoveAccountArtifacts:(NSError *)serverError
+{
+    // Removing account artifacts on invalid_grant + user_deleted_account suberror combination
+    MSIDErrorCode oauthError = MSIDErrorCodeForOAuthError(serverError.msidOauthError, MSIDErrorInternal);
```
Veena11, line 106: I see the implementation for this method is same in both subclasses, would it make sense to move it to super class ?

IdentityCore/src/requests/MSIDSilentTokenRequest.m
```diff
@@ -617,10 +619,43 @@ - (void)sendTokenRequestImpl:(MSIDRequestCompletionBlock)completionBlock
...
+                                                      clientId:requestParameters.clientId
+                                                      familyId:nil
+                                                       context:requestParameters
+                                                         error:&removalError];
+    if (!removalResult)
+    {
+        MSID_LOG_WITH_CTX_PII(MSIDLogLevelWarning, requestParameters, @"Failed to clear cache with error %@", MSID_PII_LOG_MASKABLE(removalError));
+    }
+    
+    NSError *metadataRemovalError = nil;
```
Veena11, line 650: are we proceeding with metadata removal even when account removal didnt succeed ?

Veena11 (commented):
Added few comments

IdentityCore/src/requests/sdk/adal/MSIDLegacySilentTokenRequest.m
antonioalwan: I was thinking about this when I was implementing this and and thank you for pointing that up. Left it just in case there was a need but I guess after feature completion it makes sense to move to base class since those methods match exactly

IdentityCore/src/requests/MSIDSilentTokenRequest.m
antonioalwan: True, in this case user is deleted from ESTS and all removal for any artifacts is good.

IdentityCore/src/MSIDConstants.h
```diff
@@ -215,4 +215,15 @@ extern NSString * _Nonnull const MSID_FLIGHT_USE_V2_WEB_RESPONSE_FACTORY;
...
+/**
+ * Flight to indicate if remove account artifacts should be disabled
+ * Owner: Antonio
+ * Link: N/A - No ECS flag created as this is a disable flight to be created on demand
+ * Created Date: N/A
+ * Status: Not started
+ * Fully Allocated: Not Started
+ * WorkItem: 3168316
+ */
+extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_REMOVE_ACCOUNT_ARTIFACTS;
```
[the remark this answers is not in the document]
antonioalwan: Broker and MSAL.

IdentityCore/src/requests/sdk/adal/MSIDLegacySilentTokenRequest.m
antonioalwan: Hi @Veena11 another update while implementing this:
Seems that `[super shouldRemoveAccountArtifacts:serverError];` wouldn't compile as that function is not defined in the header: MSIDSilentTokenRequest.h. This would be like a private method to be exposed in public header if I add it for subclasses to see. The fix is to have an extra header+internal. Thinking not to make more changes there for this one function call and keep the change as is.

IdentityCore/src/MSIDConstants.h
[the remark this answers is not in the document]
antonioalwan: Added test flight in Broker for MSAL: https://github.com/AzureAD/azure-activedirectory-tokenbroker-for-objc/pull/1879.

IdentityCore/src/requests/sdk/adal/MSIDLegacySilentTokenRequest.m
Veena11: I feel it is better to have the internal header file to avoid having duplicate code in multiple files ? Would the change be significant , may be I am missing something ?

Veena11 (approved):

PR #1522, opened 2025-05-14
Add method to generate jwe crypto from a give EC key

ameyapat:
## Proposed changes

Add utility method to allow MSAL/oneAuth clients to generate JWE crypto claim in payload when redeeming a bound refresh token.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/util/MSIDKeyOperationUtil.h
+++ b/IdentityCore/src/util/MSIDKeyOperationUtil.h
@@ -68,5 +68,21 @@ NS_ASSUME_NONNULL_BEGIN
                                 signingAlgorithm:(SecKeyAlgorithm)algorithm
                                          context:(_Nullable id<MSIDRequestContext>)context
                                            error:(NSError * _Nullable __autoreleasing * _Nullable)error;
+
+
+/// Generates JWE crypto using the supplied public key and algorithm.
+/// @param publicStk The public session transport key used to generate APV in jwe_crypto.
+/// @param alg The algorithm used for the purpose of the jwe_crypto. Defaults to ECDH-ES
+/// @param enc Encryption algorithm that client wants server to use to encrypt JWE repsonse. Defaults to A256GCM
+/// @param apvPrefix A prefix string attached to APV to differentiate clients.
+/// @param context request context
+/// @param error error
+- (NSDictionary *)generateJweCryptoWithTransportKey:(SecKeyRef _Nonnull)publicStk
+                                                alg:(nullable NSString *)alg
+                                                enc:(nullable NSString *)enc
+                                          apvPrefix:(nonnull NSString *)apvPrefix
+                                            context:(_Nullable id<MSIDRequestContext>)context
+                                              error:(NSError * _Nullable __autoreleasing * _Nullable)error;
+                                               
 @end
 NS_ASSUME_NONNULL_END
--- a/IdentityCore/src/util/MSIDKeyOperationUtil.m
+++ b/IdentityCore/src/util/MSIDKeyOperationUtil.m
@@ -135,6 +135,96 @@
     return signature;
 }
 
+- (nonnull NSDictionary *)generateJweCryptoWithTransportKey:(SecKeyRef _Nonnull)publicStk
+                                                        alg:(nullable NSString *)alg
+                                                        enc:(nullable NSString *)enc
+                                                  apvPrefix:(nonnull NSString *)apvPrefix
+                                                    context:(id<MSIDRequestContext> _Nullable)context
+                                                      error:(NSError * _Nullable __autoreleasing * _Nullable)error
+{
+    if (!publicStk)
+    {
+        [self generateErrorWithMessage:@"Public STK provided is not defined." underlyingError:nil context:context error:error];
+        return nil;
+    }
+    
+    NSData *exportedData = CFBridgingRelease(SecKeyCopyExternalRepresentation(publicStk, NULL));
+    if (!exportedData)
+    {
+        [self generateErrorWithMessage:@"Supplied key should be a public EC key." underlyingError:nil context:context error:error];
+        return nil;
+    }
+    
+    if (exportedData.length != 65)
+    {
+        [self generateErrorWithMessage:@"Supplied key is not a EC P-256 key." underlyingError:nil context:context error:error];
+        return nil;
+    }
+    
+    if ([NSString msidIsStringNilOrBlank:alg])
+    {
+        alg = @"ECDH-ES";
+    }
+    
+    if ([NSString msidIsStringNilOrBlank:enc])
+    {
+        enc = @"A256GCM";
+    }
+    
+    if ([NSString msidIsStringNilOrBlank:apvPrefix])
+    {
+        [self generateErrorWithMessage:@"APV prefix is not defined. A prefix must be provided to determine calling application type." underlyingError:nil context:context error:error];
+        return nil;
+    }
+    
+    /* APV is defined as:
+     <Datalen of prefix>||<prefix>||<Datalen of STK Pubkey>||<STK PubKey>||<Datalen of ASCII encoded nonce bytes>||<ASCII encoded nonce bytes>
+     prefix is string defined according to PRTv4.0 protocol for broker client or MSAL client.
+     STK Pubkey is public key of the STK.
+     nonce is client generated nonce for KDF (not AESGCM nonce)
+     => The corresponding "apv":BASE64URLEncode(OtherInfo.PartyVInfo)
+     */
+    
+    NSMutableData *data = [NSMutableData new];
+    
+    int prefixLen = (int)apvPrefix.length;
+    NSData *prefixLenData = [NSData dataWithBytes:&prefixLen length:sizeof(prefixLen)];
+    [data appendData:prefixLenData];
+    [data appendData:[apvPrefix dataUsingEncoding:NSUTF8StringEncoding]];
+    
+    
+    CFErrorRef errorRef = NULL;
+    NSData *stkData = CFBridgingRelease(SecKeyCopyExternalRepresentation(publicStk, &errorRef));
+    
+    if (!stkData)
+    {
+        [self generateErrorWithMessage:@"Failed to get data representation of public STK" underlyingError:CFBridgingRelease(errorRef) context:context error:error];
+        return nil;
+    }
+    
+    int stkLen = (int)stkData.length;
+    NSData *stkLenData = [NSData dataWithBytes:&stkLen length:sizeof(stkLen)];
+    [data appendData:stkLenData];
+    [data appendData:stkData];
+    
+    NSData *nonceData = [[NSUUID UUID].UUIDString dataUsingEncoding:NSASCIIStringEncoding];
+    int nonceLen = (int)nonceData.length;
+    NSData *nonceLenData = [NSData dataWithBytes:&nonceLen length:sizeof(nonceLen)];
+    [data appendData:nonceLenData];
+    [data appendData:nonceData];
+    
+    NSString *apvString = [data msidBase64UrlEncodedString];
+    
+    NSMutableDictionary *jweCrypto = [NSMutableDictionary new];
+    jweCrypto[@"apv"] = apvString;
+    jweCrypto[@"alg"] = alg;
+    jweCrypto[@"enc"] = enc;
+    
+    MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, nil, @"Jwe crypto generated %@", MSID_PII_LOG_MASKABLE(jweCrypto));
+    
+    return jweCrypto;
+}
+
 - (void)generateErrorWithMessage:(NSString *)errorMessage underlyingError:(NSError *)underlyingError context:(id<MSIDRequestContext> _Nullable)context error:(NSError *__autoreleasing*)error
 {
     if (error)
--- a/IdentityCore/tests/MSIDKeyOperationUtilTest.m
+++ b/IdentityCore/tests/MSIDKeyOperationUtilTest.m
@@ -216,6 +216,122 @@
     error = nil;
 }
 
+- (void)testGenerateJweCryptoWithTransportKey_validInputs_shouldReturnJweCrypto
+{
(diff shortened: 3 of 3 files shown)


IdentityCore/src/requests/MSIDJWECrypto.m
```diff
@@ -0,0 +1,92 @@
...
+        }
+        
+        if (!apv)
+        {
+            MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"JWE crypto generation failed: APV is nil");
+            MSIDFillAndLogError(error, MSIDErrorInternal, @"JWE crypto generation failed: APV is nil", context.correlationId);
+            return nil;
+        }
+        
+        if (![@"ECDH-ES" isEqual:keyExchangeAlgorithm])
```
juan-arias, line 59: nit: can you use the `MSID_JWT_ALG_ECDH` constant?

juan-arias (approved):

IdentityCore/src/requests/MSIDJWECrypto.m
ameyapat: Updated, thanks!

hieunguyenmsft (commented):
Are you going to refactor the broker code to use this class? I think that would be best as they are basically same algorithm, so in the future if anything changes for prt crypto, bound RT can just pick it up

ameyapat:
> Are you going to refactor the broker code to use this class? I think that would be best as they are basically same algorithm, so in the future if anything changes for prt crypto, bound RT can just pick it up

That is what I'm thinking when writing this code in common-core but it won't be a straight swap because if there is an issue it would have a big impact. I'll most likely generate jwe_crypto using both broker code and common-core code and compare if there are differences for requests. If there are no difference for a while, we can use the common-core code in broker

IdentityCore/src/requests/MSIDEcdhApv.m
```diff
@@ -0,0 +1,100 @@
...
+const NSUInteger kExpectedECP256KeyLength = 65;
+
+- (instancetype)initWithKey:(SecKeyRef)publicKey
+                  apvPrefix:(NSString *)apvPrefix
+                    context:(id<MSIDRequestContext> _Nullable)context
+                      error:(NSError * _Nullable __autoreleasing *)error
+{
+    if (publicKey == NULL)
+    {
+        *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal, @"Public STK provided is not defined.", nil, nil, nil, context.correlationId, nil, NO);
```
kaisong1990, line 39: Please make sure error is not nil before assigning.

IdentityCore/src/requests/MSIDEcdhApv.m
```diff
@@ -0,0 +1,100 @@
...
+        return nil;
+    }
+    
+    if ([NSString msidIsStringNilOrBlank:apvPrefix])
+    {
+        *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal, @"APV prefix is not defined. A prefix must be provided to determine calling application type.", nil, nil, nil, context.correlationId, nil, NO);
```
kaisong1990, line 45: ditto

IdentityCore/src/requests/MSIDEcdhApv.h
```diff
@@ -0,0 +1,45 @@
...
+- (instancetype)init NS_UNAVAILABLE;
++ (instancetype)new NS_UNAVAILABLE;
+
+@property (nonatomic, readonly) NSString *APV;
+@property (nonatomic, readonly) NSData *nonce;
+@property (nonatomic, readonly) NSString *apvPrefix;
+@property (nonatomic, readonly) SecKeyRef publicKey;
+
+// Format for APV: <Prefix length> | <Prefix> | <Public key length> | <Public key> | <Nonce length> | <Nonce>
+- (instancetype)initWithKey:(SecKeyRef)publicKey
```
kaisong1990, line 39: `initWithKey ` is defined within `NS_ASSUME_NONNULL_BEGIN `, but it can return nill, please consider this pattern `- (nullable instancetype)`

IdentityCore/src/requests/MSIDJWECrypto.h
```diff
@@ -0,0 +1,52 @@
...
+                   encryptionAlgorithm:(NSString *)encryptionAlgorithm
+                                   apv:(MSIDEcdhApv *)apv
+                               context:(_Nullable id<MSIDRequestContext>)context
+                                 error:(NSError * _Nullable __autoreleasing *)error;
+
+- (NSString *)urlEncodedJweCrypto;
+- (NSDictionary *)jweCryptoDictionary;
+@end
+
+@interface MSIDJWECrypto()
```
kaisong1990, line 48: Can you share some contexts on this? I have not seen this pattern before.

IdentityCore/src/requests/MSIDJWECrypto.h
```diff
@@ -0,0 +1,52 @@
...
+/// JWE response encryption algorithm
+@property (nonatomic, readonly) NSString *encryptionAlgorithm;
+/// Key exchange algorithm
+@property (nonatomic, readonly) NSString *keyExchangeAlgorithm;
+/// APV . Contains this party's public key for key exchange.
+@property (nonatomic, readonly) MSIDEcdhApv *apv;
+
+- (instancetype)init NS_UNAVAILABLE;
++ (instancetype)new NS_UNAVAILABLE;
+- (instancetype)initWithKeyExchangeAlg:(NSString *)keyExchangeAlgorithm
```
kaisong1990, line 38: same here, please consider `- (nullable instancetype)`

IdentityCore/src/requests/MSIDJWECrypto.m
```diff
@@ -0,0 +1,93 @@
...
+        
+        _keyExchangeAlgorithm = keyExchangeAlgorithm;
+        _encryptionAlgorithm = encryptionAlgorithm;
+        _apv = apv;
+        _jweCryptoDictionary = @{
+                                @"alg": keyExchangeAlgorithm,
+                                @"enc": encryptionAlgorithm,
+                                @"apv": apv.APV
+                                };
+    }
```
kaisong1990, line 82: nit: space after if block

hieunguyenmsft:
> > Are you going to refactor the broker code to use this class? I think that would be best as they are basically same algorithm, so in the future if anything changes for prt crypto, bound RT can just pick it up
> 
> That is what I'm thinking when writing this code in common-core but it won't be a straight swap because if there is an issue it would have a big impact. I'll most likely generate jwe_crypto using both broker code and common-core code and compare if there are differences for requests. If there are no difference for a while, we can use the common-core code in broker

Hmm I don't think it's a straighforward comparison, as there is nonce factor so you never able to see a value twice. Can we de-risk by a feature controlling to release on a small sample then increase? 
Anyway that is just a suggestion in a long term enhancement, not blocking this current PR. The plan for HPKE is start in Jul, and I think there will indeed new crypto flow on PRT. So the sooner we share the crypto code in common-core for both 2 procotols is better.

IdentityCore/src/requests/MSIDEcdhApv.m
ameyapat: Addressed, thanks!

IdentityCore/src/requests/MSIDEcdhApv.m
ameyapat: Addressed

IdentityCore/src/requests/MSIDEcdhApv.h
ameyapat: Addressed, thanks!

IdentityCore/src/requests/MSIDJWECrypto.h
ameyapat: Addressed, thanks!

IdentityCore/src/requests/MSIDJWECrypto.m
ameyapat: Added space

IdentityCore/src/requests/MSIDJWECrypto.h
ameyapat: This is just a way to define private properties of classes using categories. Since there is no name for the category, other classes cannot import it and hence these properties are hidden to other classes

ameyapat:
> > > Are you going to refactor the broker code to use this class? I think that would be best as they are basically same algorithm, so in the future if anything changes for prt crypto, bound RT can just pick it up
> > 
> > 
> > That is what I'm thinking when writing this code in common-core but it won't be a straight swap because if there is an issue it would have a big impact. I'll most likely generate jwe_crypto using both broker code and common-core code and compare if there are differences for requests. If there are no difference for a while, we can use the common-core code in broker
> 
> Hmm I don't think it's a straighforward comparison, as there is nonce factor so you never able to see a value twice. Can we de-risk by a feature controlling to release on a small sample then increase? Anyway that is just a suggestion in a long term enhancement, not blocking this current PR. The plan for HPKE is start in Jul, and I think there will indeed new crypto flow on PRT. So the sooner we share the crypto code in common-core for both 2 procotols is better.

Yeah, something for the future. Olga had added a customNonce argument in the broker code to put in the APV, so I can add a custom nonce and use the same one in common-core code and compare. But anyway we can think of the best strategy when the time comes

IdentityCore/src/requests/MSIDEcdhApv.m
```diff
@@ -0,0 +1,100 @@
...
+
+    CFErrorRef errorRef = NULL;
+    CFDictionaryRef attributes = SecKeyCopyAttributes(publicKey);
+    CFStringRef keyType = CFDictionaryGetValue(attributes, kSecAttrKeyType);
+    CFStringRef keyClass = CFDictionaryGetValue(attributes, kSecAttrKeyClass);
+
+    BOOL isECPrivateKey = (keyType == kSecAttrKeyTypeECSECPrimeRandom) && (keyClass == kSecAttrKeyClassPrivate);
+
+    if (attributes) CFRelease(attributes);
+    NSData *stkData = CFBridgingRelease(SecKeyCopyExternalRepresentation(publicKey, NULL));
```
kaisong1990, line 57: Please consider log the error instead of silent it in `SecKeyCopyExternalRepresentation`

IdentityCore/src/requests/MSIDJWECrypto.h
```diff
@@ -0,0 +1,52 @@
...
+
+- (instancetype)init NS_UNAVAILABLE;
++ (instancetype)new NS_UNAVAILABLE;
+- (nullable instancetype)initWithKeyExchangeAlg:(NSString *)keyExchangeAlgorithm
+                            encryptionAlgorithm:(NSString *)encryptionAlgorithm
+                                            apv:(MSIDEcdhApv *)apv
+                                        context:(_Nullable id<MSIDRequestContext>)context
+                                          error:(NSError * _Nullable __autoreleasing *)error;
+
+- (NSString *)urlEncodedJweCrypto;
```
kaisong1990, line 44: Since this is a lazy loading, Is this thread safe?

IdentityCore/src/requests/MSIDJWECrypto.m
```diff
@@ -0,0 +1,94 @@
...
+            return nil;
+        }
+        
+        _keyExchangeAlgorithm = keyExchangeAlgorithm;
+        _encryptionAlgorithm = encryptionAlgorithm;
+        _apv = apv;
+        _jweCryptoDictionary = @{
+                                @"alg": keyExchangeAlgorithm,
+                                @"enc": encryptionAlgorithm,
+                                @"apv": apv.APV
```
kaisong1990, line 80: please consider making them const

IdentityCore/src/requests/MSIDJWECrypto.h
kaisong1990: If this is the private class, It is better to move this part into `.m` file instead of the same `.h` file

kaisong1990:
Rest looks good, please take a look new comment

IdentityCore/src/requests/MSIDJWECrypto.h
ameyapat: It is a readonly property and I don't anticipate multiple threads contending to read this. Thread safety is not required here

IdentityCore/src/requests/MSIDJWECrypto.m
ameyapat: Added constants

ameyapat:
https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1522/#discussion_r2101133461
Shifted to .m

IdentityCore/src/requests/MSIDJWECrypto.m
```diff
@@ -0,0 +1,99 @@
...
+        {
+            MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"JWE crypto generation failed: Key exchange algorithm is nil or blank");
+            MSIDFillAndLogError(error, MSIDErrorInternal, @"JWE crypto generation failed: Key exchange algorithm is nil or blank", context.correlationId);
+            return nil;
+        }
+        
+        if ([NSString msidIsStringNilOrBlank:encryptionAlgorithm])
+        {
+            MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"JWE crypto generation failed: Encryption algorithm is nil or blank");
+            MSIDFillAndLogError(error, MSIDErrorInternal, @"JWE crypto generation failed: Encryption algorithm is nil or blank", context.correlationId);
```
antrix1989, line 54: Should we just keep MSIDFillAndLogError? it will log as well, otherwise we duplicate logs.

IdentityCore/src/requests/MSIDEcdhApv.m
kaisong1990: Not necessarily, if we can check where causes the failure, should be good

kaisong1990 (approved):

juan-arias (approved):

PR #1529, opened 2025-05-22
Disable preferred identity in CBA flow

hieunguyenmsft:
## Proposed changes

Disable preferred identity save/query in CBA flow to let user choose cert.

```
By design, CBA supports multiple accounts with different certificates on the same device.
The hostname (certauth.login.microsoftonline.com) and distinguished names (list of subject names of all the cert authorities configured in the tenant) are used to find the correct certificate for authentication. There is no user-specific unique ID per certificate because, in CBA, a user can have multiple certificates to authenticate the same UPN, and a certificate can be used to authenticate multiple users (e.g., personal/admin account, etc.).
The way it works relies on the user deciding the correct certificate, so the client needs to present the certificate picker.
The cache problem: We have two cache layers here: internal browser (including embedded webview that our libraries start for interactive requests) and keychain record that common-core/broker owns.
Browser cache saves the certificate picked from the previous session, and next time, if the same challenge and conditions (hostname, distinguished name) are met, it will be sent to the server without showing a prompt.
Keychain cache is permanent, and common-core will prioritize the “preferred identity” instead of showing the prompt to the user. (code snippet at the bottom)
So, in order for the user to choose another certificate, both caches need to be flushed. 

```## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

### DIFF
--- a/IdentityCore/src/webview/embeddedWebview/challangeHandlers/mac/MSIDCertAuthHandler.m
+++ b/IdentityCore/src/webview/embeddedWebview/challangeHandlers/mac/MSIDCertAuthHandler.m
@@ -107,7 +107,7 @@
 
 + (BOOL)isIdentityPersistenceEnabled
 {
-    return [MSIDFlightManager.sharedInstance boolForKey:MSID_FLIGHT_DISABLE_PREFERRED_IDENTITY_CBA];
+    return ![MSIDFlightManager.sharedInstance boolForKey:MSID_FLIGHT_DISABLE_PREFERRED_IDENTITY_CBA];
 }
 
 + (void)respondCertAuthChallengeWithIdentity:(nonnull SecIdentityRef)identity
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: IdentityCore/src/webview/embeddedWebview/challangeHandlers/mac/MSIDCertAuthHandler.m [hunk 1]
```diff
@@ -40,20 +42,23 @@ + (BOOL)handleChallenge:(NSURLAuthenticationChallenge *)challenge
 {
     NSString *host = challenge.protectionSpace.host;
     NSArray<NSData*> *distinguishedNames = challenge.protectionSpace.distinguishedNames;
-    
-    // Check if a preferred identity is set for this host
-    SecIdentityRef identity = SecIdentityCopyPreferred((CFStringRef)host, NULL, (CFArrayRef)distinguishedNames);
-    
-    if (!identity)
+    BOOL isIdentityValid = false;
+    SecIdentityRef identity = NULL;
```
kaisong1990, line 46: Do you need o release `identity`?

PR #1534, opened 2025-06-17
Added ephemeral browser session to duna

josephpab:
## Proposed changes

When ESTS redirects back to MSAL at /switch_browser, it will add a new query parameter called browser_modes. This parameter is defined as a base64 encoded int that can be extended in the future with multiple options. The integer is the sum of multiple flags that can be summed together to indicate one or more options. The flag PrivateSession = 1 << 0 which if present indicates that the system browser should be opened in a private session/ephemeral mode if supported. 

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

### DIFF
--- a/IdentityCore/src/webview/operations/ios/MSIDSwitchBrowserOperation.m
+++ b/IdentityCore/src/webview/operations/ios/MSIDSwitchBrowserOperation.m
@@ -86,10 +86,17 @@
     NSURLComponents *requestURLComponents = [[NSURLComponents alloc] initWithString:self.switchBrowserResponse.actionUri];
     requestURLComponents.percentEncodedQuery = [queryItems msidURLEncode];
     NSURL *startURL = requestURLComponents.URL;
+    BOOL usePrivateSession = YES;
+
+    if (self.switchBrowserResponse.bitMask)
+    {
+        usePrivateSession = (self.switchBrowserResponse.bitMask & (1 << 0)) != 0;
+    }
     
     [self.certAuthManager startWithURL:startURL
                       parentController:requestParameters.parentViewController
                                context:requestParameters
+            ephemeralWebBrowserSession:usePrivateSession
                        completionBlock:^(NSURL *callbackURL, NSError *error)
      {
         [self.certAuthManager resetState];
--- a/IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.h
+++ b/IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.h
@@ -29,6 +29,7 @@
 
 @property (nonatomic, readonly) NSString *actionUri;
 @property (nonatomic, readonly) NSString *switchBrowserSessionToken;
+@property (nonatomic, readonly) NSInteger bitMask;
 
 - (instancetype )init NS_UNAVAILABLE;
 + (instancetype)new NS_UNAVAILABLE;
--- a/IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.m
+++ b/IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.m
@@ -49,6 +49,8 @@
         if (![self isMyUrl:url redirectUri:redirectUri]) return nil;
         
         _actionUri = self.parameters[@"action_uri"];
+        _bitMask = [self.parameters[@"browser_modes"] integerValue];
+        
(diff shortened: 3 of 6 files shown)


IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.m
```diff
@@ -47,8 +47,28 @@ - (instancetype)initWithURL:(NSURL *)url
...
+            NSInteger bitmask = 0;
+            
+            if (decoded.length > 0)
+            {
+                const uint8_t *bytes = decoded.bytes;
+                for (NSUInteger i = 0; i < decoded.length; i++)
+                {
+                    bitmask = (bitmask << 8) | bytes[i];
+                }
+
```
antonioalwan, line 66: Trying to understand this logic:
- you're operating on a byte `bytes = decoded.bytes` however the for-loop is checking the `decoded.length`. So if decoded.length is 20. I assume it is 20 bytes?
- then you're taking each byte and running the check for 1st bit.
In this flow I see:
1. How many  bytes should we expect, that should be identified?
2. If one byte is 1 for the bitmask and the one after it is not, then your value is wrong and which one is correct?
3. I think if you make the expectation clear by expecting 4 bytes (integer). and take the byte that should be expected set and operate on it. Then it would be more concise.

IdentityCore/src/webview/operations/ios/MSIDSwitchBrowserOperation.m
```diff
@@ -86,10 +86,17 @@ - (void)invokeWithRequestParameters:(nonnull MSIDInteractiveTokenRequestParamete
     NSURLComponents *requestURLComponents = [[NSURLComponents alloc] initWithString:self.switchBrowserResponse.actionUri];
     requestURLComponents.percentEncodedQuery = [queryItems msidURLEncode];
     NSURL *startURL = requestURLComponents.URL;
+    BOOL usePrivateSession = YES;
```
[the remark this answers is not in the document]
josephpab: No, before my change, we would always pass YES to ephemeralWebBrowserSession. We want to continue with default behavior unless browser_modes parameter says otherwise

IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.m
josephpab: Logic changed, should be better now

IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.h
```diff
@@ -27,8 +27,15 @@
...
 
+typedef NS_OPTIONS(NSInteger, MSIDSwitchBrowserModes) {
+    BrowserModePrivateSession = 1 << 0,
+    // Add future flags here
+};
+
 @property (nonatomic, readonly) NSString *actionUri;
 @property (nonatomic, readonly) NSString *switchBrowserSessionToken;
+@property (nonatomic, readonly) NSInteger bitMask;
+@property (nonatomic, readonly) BOOL hasBitMask;
```
antonioalwan, line 38: nit: bitMask doesn't mean a lot. Just call those vars as browserModesBitMask or similar.

IdentityCore/src/webview/response/MSIDSwitchBrowserResponse.m
```diff
@@ -47,8 +48,21 @@ - (instancetype)initWithURL:(NSURL *)url
...
         _actionUri = self.parameters[@"action_uri"];
+        
+        NSString* browserOptionsString = self.parameters[@"browser_modes"];
+        if (browserOptionsString)
+        {
+            NSData *data = [NSData msidDataFromBase64UrlEncodedString:browserOptionsString];
+            uint32_t flagsValue = 0;
+            [data getBytes:&flagsValue length:sizeof(flagsValue)];
+
+            _bitMask = flagsValue;
```
antonioalwan, line 62: I would have preferred that handling the browserModes happens here and you have an answer here transformed into one BOOL variable instead of splitting this logic into 2 files.

josephpab: Done

PR #1655, opened 2026-01-14
[major][feature]: Add request execution flow logger

kaisong1990:
## Proposed changes

[Design doc](https://dev.azure.com/IdentityDivision/DevEx/_git/AuthLibrariesApiReview?version=GB3446882/brk-cc-perf-tag&path=/%5BApple%5D%20broker_common_perf_tag/broker_common_perf_tag.md&_a=preview)

This pull request introduces a new "execution flow" logging and tracing subsystem to the project, along with related unit tests and integration into the `MSIDHttpRequest` workflow. The changes primarily add new source and header files, update the Xcode project to include these files in the appropriate build phases and groups, and add logging hooks to the HTTP request code.

The most important changes are:

**Execution Flow Subsystem Addition:**
[...]

PR #1699, opened 2026-02-19
[MINOR] [Feature]: New Contract Between ESTS and Broker Meta Data

agubuzomaximus:
## Proposed changes

Captures the x-ms-clientData HTTP response header from the /token endpoint, flows it through MSIDTokenResponse into MSIDTokenResult.brokerMetaData, enabling downstream consumers (broker telemetry, OneAuth) to access server-side diagnostic data.

This is phase one and only coveres the happy path. 

see: 
https://microsoft-my.sharepoint-df.com/:w:/p/kasong/cQotl0mr8ADKTItsegyyctU0EgUCJ2uEuEskcIy-XbnUCkuovw
https://microsoft.sharepoint-df.com/:w:/s/AuthSTSDocs/cQrIrGUMsdxVSZE3GlsMkqaBEgUCuB5LueV1cawNVF1-Yt3n0w

## Type of change

- [X] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


galwarsamy (approved):

kaisong1990 (commented):
Looks good overall, please add test as suggested. will you have a separate PR to address the unhappy path from both /auth and /token endpoint?

kaisong1990 (approved):

IdentityCore/src/requests/MSIDTokenResponseHandler.m
```diff
@@ -112,6 +113,12 @@ - (void)handleTokenResponse:(MSIDTokenResponse *)tokenResponse
 
     [tokenResult insertBrokerMetaData:@(responseLatency) forKey:MSID_TOKEN_RESULT_BROKER_APP_RESPONSE_LATENCY];
     [tokenResult insertBrokerMetaData:@(brokerHandlingTime) forKey:MSID_TOKEN_RESULT_BROKER_APP_BROKER_HANDLING_TIME_INTERVAL];
+
+    if (tokenResponse.clientData)
+    {
+        [tokenResult insertBrokerMetaData:tokenResponse.clientData forKey:MSID_TOKEN_RESULT_CLIENT_DATA];
```
kaisong1990: This part is not triggerred when token request happens in broker and you might need to add similar function into `MSIDTokenResponseValidator` when create `MSIDTokenResult`  (~L121), so that client data can propagate.

kaisong1990 (changes requested):

kaisong1990 (commented):
~test comment to check auth~

kaisong1990 (commented):
## Code Review

Hi @agubuzomaximus please check if those AI reviews makes sense:


Reviewed as an Apple senior engineer working on iOS/macOS authentication libraries. Below are the issues found, ordered by severity.

---

### 🔴 Bug: `MSID_CLIENT_DATA_RESPONSE` missing from `knownFields` in `MSIDAADTokenResponse.setAdditionalServerInfo:`

**File:** `IdentityCore/src/oauth2/aad_base/MSIDAADTokenResponse.m`

The `setAdditionalServerInfo:` override exists specifically to strip individually-parsed fields from the catch-all `additionalServerInfo` dict — this is exactly how `MSID_CCS_REQUEST_ID_RESPONSE` and `MSID_CCS_REQUEST_SEQUENCE_RESPONSE` are handled. But `MSID_CLIENT_DATA_RESPONSE` was not added to that list.

Because the preprocessor injects `client-data` into the raw JSON dict and the base class deserializes the entire dict into `additionalServerInfo` before `initWithJSONDictionary` reads it out into `self.clientData`, the value ends up stored in **both places**. `jsonDictionary` then writes it a second time via `json[MSID_CLIENT_DATA_RESPONSE] = self.clientData`, so any round-trip serialization will double the field.

Fix — same pattern as the CCS fields:
```objc
NSArray *knownFields = @[...,
                         MSID_CCS_REQUEST_ID_RESPONSE,
                         MSID_CCS_REQUEST_SEQUENCE_RESPONSE,
                         MSID_CLIENT_DATA_RESPONSE]; // add this
```

A test verifying `additionalServerInfo` does **not** contain `client-data` when the header is present is also needed.

---

### 🟡 Commented-out debug code should not be committed to production source

Three separate files ship with commented-out debug/scaffolding blocks that should be cleaned up before merge:

1. **`MSIDWebOAuth2Response.m` lines 72–83** — large `#if DEBUG` block that artificially injects a synthetic `/authorize` error. This is clearly test scaffolding and is confusing to future readers.

2. **`MSIDWebviewFactory.m` lines 221–224** — commented-out `dc` query parameter for targeting a test slice. This belongs in a local `.gitignore`d file, not production source.

3. **`MSIDRequestParameters.m` lines 167–169** — same commented-out `dc` QP on the token endpoint.

All three should be deleted entirely before merging.

---

### 🟡 `clidata=1` is unconditionally appended to every `/authorize` request

**File:** `IdentityCore/src/oauth2/MSIDWebviewFactory.m`, line 227

The comment says this flag enables STS to return client-data on `/authorize` errors. However, this is placed in the **base** `MSIDWebviewFactory` so it applies to every MSAL interactive flow — not just broker-aware ones. Non-broker clients will begin advertising this flag but have no consumer for the data.

Is this intentional? If only broker-aware callers need this, it should be conditioned on whether the request is going through the broker path. If it is intentional (i.e. we want all clients to opt in), please add a comment explaining that rationale.

---

### 🟡 Broker path (`validateAndSaveBrokerResponse`) has no test coverage for `clientData` propagation

**File:** `IdentityCore/src/requests/sdk/MSIDTokenResponseValidator.m`

The `clientData` insertion happens inside `validateTokenResponse:`, which is called by **both** `validateAndSaveTokenResponse:` and `validateAndSaveBrokerResponse:`. The new tests (`testValidateTokenResponse_whenClientDataIsPresent_*`) cover the direct path, but the broker response path (`validateAndSaveBrokerResponse:`) is not exercised. Given this is a broker metadata contract change, a test through that path would be valuable.

---

### 🟢 Nit: hardcoded string literal `@"clidata"` in test instead of the constant

**File:** `IdentityCore/tests/MSIDWebOAuth2ResponseTests.m`, line 139

```objc
@"clidata" : clientData,  // should be MSID_OAUTH2_CLIENT_DATA_QUERY_PARAM
```

---

### 🟢 Nit: trailing whitespace introduced

**File:** `IdentityCore/src/webview/embeddedWebview/MSIDOAuth2EmbeddedWebviewController.m`

The diff adds trailing whitespace after `{` on the `initWithStartURL:` opening brace line. This is cosmetic but will fail any whitespace-checking CI hooks.

---

### 🟢 Consider a dedicated broker metadata key constant for `clientData`

**File:** `IdentityCore/src/requests/sdk/MSIDTokenResponseValidator.m`, line 80

```objc
[tokenResult insertBrokerMetaData:tokenResponse.clientData forKey:MSID_CLIENT_DATA_RESPONSE];
```

`MSID_CLIENT_DATA_RESPONSE` (`@"client-data"`) is also used as the JSON serialisation key throughout the token response. Reusing it as the `brokerMetaData` dictionary key means consumers cannot distinguish the serialisation context from the broker context. The existing pattern uses dedicated constants like `MSID_TOKEN_RESULT_BROKER_APP_VERSION`. Consider defining `MSID_TOKEN_RESULT_CLIENT_DATA` for the `brokerMetaData` key, similar to how other broker metadata keys are named.

IdentityCore/tests/MSIDWebOAuth2ResponseTests.m
```diff
@@ -120,6 +120,33 @@ - (void)testInitWithParameters_whenNoRequestStateAndOAuthServerErrorWithValidSta
...
+    NSString *errorDescription = @"error description";
+    NSString *subError = @"suberror";
+    NSString *clientData = @"1|50076|basic_action||";
+
+    NSURLComponents *urlComponents = [NSURLComponents componentsWithString:@"https://contoso.com"];
+    urlComponents.queryItems = @{
+                                 MSID_OAUTH2_ERROR : errorString,
+                                 MSID_OAUTH2_ERROR_DESCRIPTION : errorDescription,
+                                 MSID_OAUTH2_SUB_ERROR : subError,
+                                 @"clidata" : clientData,
```
kaisong1990: Nit: use the constant instead of the raw string literal.

`@"clidata"` should be `MSID_OAUTH2_CLIENT_DATA_QUERY_PARAM`. The constant is already imported in nearby tests — using it here keeps the tests resilient to value changes and makes the intent self-documenting.

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -217,6 +217,14 @@ @implementation MSIDWebviewFactory
...
     
     [result addEntriesFromDictionary:[self metadataFromRequestParameters:parameters]];
+
+//    // When in debug mode add test slice from STS to allow MSAL to obtain error info telemetry.
+//    #if DEBUG
+//         result[@"dc"] = @"ESTS-PUB-SCUS-FD000-TEST1-100";
+//    #endif
+    
+    // Include client telemetry data request flag so STS can return client-data on /authorize errors.
+    result[MSID_OAUTH2_CLIENT_DATA_QUERY_PARAM] = @"1";
```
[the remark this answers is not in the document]
kaisong1990: If this is only for testsing purpose, please ignore

IdentityCore/src/webview/response/MSIDWebOAuth2Response.m
```diff
@@ -67,18 +68,34 @@ - (NSError *)oauthErrorFromParameters:(NSDictionary *)parameters
...
+//    NSMutableDictionary *mutableParameters = [parameters mutableCopy];
+//    if (!mutableParameters[MSID_OAUTH2_ERROR])
+//    {
+//        mutableParameters[MSID_OAUTH2_ERROR] = @"interaction_required";
+//        mutableParameters[MSID_OAUTH2_ERROR_DESCRIPTION] = @"[DEBUG] Mocked STS error for clidata flow testing";
+//        mutableParameters[MSID_OAUTH2_SUB_ERROR] = @"basic_action";
+//        mutableParameters[MSID_OAUTH2_CLIENT_DATA_QUERY_PARAM] = @"mock_clidata_value|50076|basic_action||";
+//    }
+//    parameters = mutableParameters;
+//#endif
```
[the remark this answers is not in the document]
kaisong1990: If this is only for testsing purpose, please ignore

kaisong1990 (commented):
Overall the approach is solid. Capturing `x-ms-clientData` from the /token response header and `clidata` from the /authorize redirect, flowing both through to `brokerMetaData` and error `userInfo`, makes sense. Test coverage is good across the different code paths.

The `knownFields` addition in `MSIDAADTokenResponse.setAdditionalServerInfo:` correctly prevents clientData from leaking into `additionalServerInfo`. The dual-path coverage (base validator + legacy validator) is correct since the legacy override completely replaces the base `createTokenResultFromResponse:`. Error enrichment in both network-level and validation-level code paths covers the right scenarios.

A few inline comments below. The `MSIDWebviewFactory.m` change is whitespace-only (trailing space removal) — consider dropping it to keep the diff focused.

IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -163,6 +163,7 @@ - (NSURL *)tokenEndpoint
     {
         [endpointQPs addEntriesFromDictionary:self.extraURLQueryParameters];
     }
+
```
kaisong1990: This adds a stray blank line, resulting in a double blank line here. Looks accidental — please revert.

IdentityCore/src/requests/MSIDTokenResponseHandler.m
```diff
@@ -28,6 +28,7 @@
 #import "MSIDAccountIdentifier.h"
 #import "MSIDTokenResult.h"
 #import "MSIDAccount.h"
+#import "MSIDOAuth2Constants.h"
```
kaisong1990: This import doesn't appear to be used anywhere in this file. The clientData insertion happens in `MSIDTokenResponseValidator` and `MSIDLegacyTokenResponseValidator`, not here. Remove the unused import.

IdentityCore/src/network/MSIDHttpRequest.m
```diff
@@ -155,6 +156,15 @@ - (void)sendWithBlock:(MSIDHttpRequestDidCompleteBlock)completionBlock
 
           if (error)
           {
+              NSString *clientData = httpResponse.allHeaderFields[MSID_CLIENT_DATA_HEADER_KEY];
+              if (clientData)
+              {
+                  MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.context, @"Error in MSIDHTTPRequest SendWithBlock and client data is available, Error should be stored in user info and passed up to MSAL via callbacks.");
```
kaisong1990: This log message reads more like a design comment than a runtime log. Callers reading logs shouldn't see prescriptive statements about what "should" happen. Something like `@"Enriching error userInfo with client data from response header."` would be more appropriate.

IdentityCore/tests/MSIDWebOAuth2ResponseTests.m
kaisong1990: Use the constant `MSID_OAUTH2_CLIENT_DATA_QUERY_PARAM` instead of the raw string `@"clidata"`. The constant is already imported in this file — using it keeps the test resilient to value changes.

IdentityCore/src/webview/embeddedWebview/MSIDOAuth2EmbeddedWebviewController.m
```diff
@@ -219,6 +219,7 @@ - (void)endWebAuthWithURL:(NSURL *)endURL
     
     if (error)
     {
+        // TODO: check if we can pass the STS info back from here too
```
kaisong1990: Fine for phase 1, but please either reference a work item here or file one so this doesn't get lost.

changelog.txt
```diff
@@ -1,5 +1,7 @@
 TBD
 * Add retry logic when bound app refresh token redemption fails.
+* Add token request execution flow logger (#1655)
```
kaisong1990: This entry (`Add token request execution flow logger (#1655)`) already exists under `Version 1.20.0` on `dev`. Adding it here under `TBD` will create a duplicate when merged. Remove this line and keep only the #1699 entry.

IdentityCore/src/requests/sdk/msal/MSIDDefaultBrokerResponseHandler.m
```diff
@@ -59,6 +59,7 @@ - (instancetype)initWithOauthFactory:(MSIDOauth2Factory *)factory
                                 @"home_account_id" : MSIDHomeAccountIdkey,
                                 @"declined_scopes" : MSIDDeclinedScopesKey,
                                 @"granted_scopes" : MSIDGrantedScopesKey,
+                                @"client-data" : MSID_CLIENT_DATA_RESPONSE,
```
kaisong1990: `MSID_CLIENT_DATA_RESPONSE` resolves to `@"client-data"`, making this a key-to-itself mapping. It works because the mapping dictionary is used to filter which keys get promoted into the error `userInfo`, but if someone renames the constant value later this would silently break. Not blocking, just flagging for awareness.

kaisong1990:
I will also do a quick run tomorrow to make sure the data is populated into broker response

kaisong1990 (approved):

kaisong1990 (approved):

PR #1807, opened 2026-05-07
[minor][engg]: Add MSIDDIContainer dependency-injection container

kaisong1990:
## Summary

Adds **`MSIDDIContainer`** — a lightweight, thread-safe dependency-injection container under `IdentityCore/src/util/`. It lets producers register a factory keyed by class or protocol, and lets consumers resolve a concrete instance for that key without taking a direct dependency on the producer. Tests can install per-key overrides to inject mocks **without method swizzling**.

## Motivation

The downstream broker repo currently relies heavily on `MSIDTestSwizzle` to swap `+[ADBrokerFlightManager sharedInstance]` for a mock in unit tests. Swizzling is fragile and leaks state across tests. Standing the container up here in CommonCore means broker, MSAL and any future Microsoft Identity client can adopt the same DI seam.

This PR only adds the container + its tests. **No existing CommonCore code is modified to consume it** — adoption happens in downstream PRs (broker first).

## API

\`\`\`objc
typedef NS_ENUM(NSInteger, MSIDDIContainerLifetime) {
    MSIDDIContainerLifetimeSingleton,
    MSIDDIContainerLifetimeTransient
};

@interface MSIDDIContainer : NSObject
@property (class, nonatomic, readonly) MSIDDIContainer *sharedInstance;

- (void)registerClass:(Class)cls lifetime:(MSIDDIContainerLifetime)lifetime factory:(id (^)(void))factory;
- (void)registerProtocol:(Protocol *)proto lifetime:(MSIDDIContainerLifetime)lifetime factory:(id (^)(void))factory;

- (id)resolveClass:(Class)cls;
- (id)resolveProtocol:(Protocol *)proto;

- (void)setOverrideForClass:(Class)cls instance:(id)instance;
- (void)setOverrideForProtocol:(Protocol *)proto instance:(id)instance;
- (void)removeOverrideForClass:(Class)cls;
- (void)removeOverrideForProtocol:(Protocol *)proto;
- (void)resetAllOverrides;
@end
\`\`\`

Resolution order for any key: **override → cached singleton (if applicable) → registered factory**.
Unregistered key with no override → \`NSAssert\` (debug) + \`@throw NSInternalInconsistencyException\` (release).

## Tests

10 unit tests in \`MSIDDIContainerTests.m\` cover:

- Singleton lifetime returns identical instance across resolves
- Transient lifetime returns a fresh instance per resolve
- Protocol-keyed register + resolve
- Override precedence over factory and over cached singleton
- \`resetAllOverrides\` restores factory behavior
- \`removeOverrideForClass:\` clears only that key
- Resolving an unregistered key throws \`NSInternalInconsistencyException\`
- Concurrent resolution via \`dispatch_apply\` (200 iterations) returns the same singleton instance
- \`+sharedInstance\` is stable

All 10 pass on iOS (iPhone 16 Pro / iOS 18.6 simulator). Mac scheme also builds cleanly.

## Risk

- Pure addition — no existing call sites changed.
- Header is added with default Project visibility (matches \`MSIDFlightManager.h\` etc.); broker/MSAL consume CommonCore as source so direct \`#import "MSIDDIContainer.h"\` works through the same header search paths.
- Container is opt-in: nothing instantiates it until a consumer calls \`+sharedInstance\` or \`registerClass:…\`.

## Checklist

- [x] Code compiles for iOS and Mac targets
- [x] New unit tests added and passing locally
- [x] No public API changed; no deprecations
- [x] Follows IdentityCore code style (4-space indent, Allman braces, \`MSID\` prefix, properties over ivars)
- [x] License header on every new file


kaisong1990:
Thanks for the review! Pushed 6da1dd589 addressing all six findings:

1. **changelog.txt** — added entry for #1807 at top of TBD section.
2. **Singleton lifetime race** — `resolveKey:` now invokes the factory **while holding** `_lock` for `MSIDDIContainerLifetimeSingleton`, guaranteeing exactly-once initialization even under contention. Trade-off (factories must be cheap and not re-enter the container for the same key) is documented inline. Transient lifetime still invokes the factory outside the lock.
3. **`NSAssert` + nil factory return** — `NSAssert` becomes a no-op in release builds, and for singletons `_singletonCache[key] = nil` would silently delete the registration. Replaced with explicit nil check that throws `NSInternalInconsistencyException` for both lifetimes (matches the unregistered-key path).
4. **Test fixture init brace style** — single-line `if (self) _identifier = ...;` rewritten to multi-line Allman with braces.
5. **Test fixture method same-line braces** — `{ return @"hello"; }` expanded to Allman.
6. **Concurrency test** — strengthened from `XCTAssertGreaterThanOrEqual(factoryInvocations, 1)` to `XCTAssertEqual(factoryInvocations, 1)` now that the lock-held init guarantees exactly one invocation. Also added `testResolveClass_whenFactoryReturnsNil_shouldThrow`.

All 11 `MSIDDIContainerTests` pass locally on iOS (iPhone 16 Pro / iOS 18.6) and macOS.

kaisong1990:
Added `resolveClass:orDefault:` / `resolveProtocol:orDefault:` in d6044ecf7 per downstream consumer feedback.

**Why:** lets a class's existing `+sharedInstance` funnel through the container with **no bootstrap step required**. Production behavior is identical to today (caller's own `dispatch_once` returns the real instance); tests still call `setOverrideForClass:instance:` to inject mocks.

**Adoption pattern (broker example):**
```objc
+ (instancetype)sharedInstance
{
    static dispatch_once_t onceToken;
    static ADBrokerFlightManager *defaultInstance = nil;
    return [[MSIDDIContainer sharedInstance]
            resolveClass:self
            orDefault:^id {
                dispatch_once(&onceToken, ^{
                    defaultInstance = [[ADBrokerFlightManager alloc] init];
                });
                return defaultInstance;
            }];
}
```

Resolution order: override → cached singleton → registered factory → default provider. The default-provider path **does not cache** — caller owns its own singleton storage, so lifetime semantics stay predictable. Default block returning nil throws `NSInternalInconsistencyException`, matching the unregistered-key path.

6 new tests (17 total). All pass on iOS (iPhone 16 Pro / iOS 18.6) and macOS.

kaisong1990:
Added typed `Class<Protocol>` API for class-method seams (`d65c66968`).

**Why:** The container already supports overriding class methods via the Class-as-id pattern (`setOverrideForClass:instance:(id)[FakeClass class]`), but the call site loses static type checking. These typed wrappers let callers declare a protocol with `+` class methods, have production and fake classes conform to it, and resolve a `Class<Protocol>` — so the compiler verifies every override implements the protocol's class methods.

**API added** (1-line forwarders over the existing `_overrideByKey` storage; no new state):
- `resolveImplClassForProtocol:orDefault:` → `Class<Protocol>`
- `resolveImplClassForClass:orDefault:` → `Class`
- `setImplClassOverride:forProtocol:`
- `setImplClassOverride:forClass:`

**Migration recipe** for an existing class-method swizzle (e.g., `+[MSIDWorkPlaceJoinUtil getWPJKeysWithTenantId:context:]`):

```objc
@protocol MSIDWPJKeysProviding <NSObject>
+ (nullable MSIDWPJKeyPairWithCert *)getWPJKeysWithTenantId:(NSString *)t context:(id)c;
@end

@interface MSIDWorkPlaceJoinUtil (DI) <MSIDWPJKeysProviding> @end

// Production funnel:
+ (MSIDWPJKeyPairWithCert *)getWPJKeysWithTenantId:(NSString *)t context:(id)c
{
    Class<MSIDWPJKeysProviding> impl =
        [[MSIDDIContainer sharedInstance]
            resolveImplClassForProtocol:@protocol(MSIDWPJKeysProviding)
                              orDefault:^Class { return self; }];
    if (impl == self) return [self _msidDefault_getWPJKeysWithTenantId:t context:c];
    return [impl getWPJKeysWithTenantId:t context:c];
}

// Test:
@interface MSIDFakeWPJKeysProvider : NSObject <MSIDWPJKeysProviding> @end
[[MSIDDIContainer sharedInstance]
    setImplClassOverride:[MSIDFakeWPJKeysProvider class]
             forProtocol:@protocol(MSIDWPJKeysProviding)];
// tearDown: [[MSIDDIContainer sharedInstance] resetAllOverrides];
```

**Coverage:** 6 new tests (23 total), all green on iOS 18.6 sim and macOS.

kaisong1990:
Please refer to this link on how to use it in both UTs and current production code:

https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/commit/55cef3ce1457731c2b846992bfda573655efc00b

IdentityCore/src/util/MSIDDIContainer.m
```diff
@@ -0,0 +1,348 @@
...
+- (instancetype)init
+{
+    self = [super init];
+
+    if (self)
+    {
+        _entryByKey = [NSMutableDictionary new];
+        _overrideByKey = [NSMutableDictionary new];
+        _singletonCache = [NSMutableDictionary new];
+        _lock = [NSLock new];
```
antrix1989: lets use dispatch_barrier instead of NSLock, see MSIDCache for reference. We don't need to block when we read, only when we write.

IdentityCore/src/util/MSIDDIContainer.h
```diff
@@ -0,0 +1,285 @@
...
+ * when the seam is a class method on a concrete utility class and you do
+ * not want to introduce a separate protocol — the override class need only
+ * supply matching @c + selectors at runtime.
+ *
+ * @see resolveImplClassForProtocol:orDefault:
+ */
+- (Class)resolveImplClassForClass:(Class)cls
+                        orDefault:(Class _Nonnull (^)(void))defaultProvider;
+
+#pragma mark - Test overrides
```
antrix1989: do we actually need to have "override" methods for tests only? Why don't we use real prod methods to set mocks?

IdentityCore/src/util/MSIDDIContainer.m
kaisong1990: Great call — switched to a concurrent dispatch_queue + dispatch_barrier in `ed65346ce`, mirroring `MSIDCache.m`:

- **Reads** (`resolveKey:`, the hot path that every consumer hits on every call) now run through `dispatch_sync` on a concurrent queue, so multiple resolves can proceed in parallel.
- **Writes** (`registerKey:`, `setOverrideForClass:instance:`, `setOverrideForProtocol:instance:`, `setImplClassOverride:forProtocol:`, `setImplClassOverride:forClass:`, `removeOverrideFor*`, `resetAllOverrides`) take the queue with `dispatch_barrier_sync`.
- **Singleton install** is now a double-checked write inside a barrier (re-check override → re-check cache → invoke factory → install) so the exactly-once guarantee is preserved without serializing reads.

The `_lock` ivar is gone; replaced with `_synchronizationQueue` named `com.microsoft.msiddicontainer-<UUID>` to match the cache's pattern.

All 23 `MSIDDIContainerTests` still pass on iOS 18.6 sim, including `testResolveClass_whenInvokedConcurrently_shouldRunFactoryOnlyOnce` (`XCTAssertEqual(factoryInvocations, 1)`).

IdentityCore/src/util/MSIDDIContainer.h
kaisong1990: Done in 67aead9c9 — removed all `setOverrideFor*` / `setImplClassOverride*` / `removeOverrideFor*` / `resetAllOverrides` APIs. Tests now register mocks through the same `register*:lifetime:factory:` APIs production code uses (a shared base test class' `setUp` is the natural place for cross-test seams), and call the new `-reset` from `tearDown` to leave the container clean. Net API surface: -7 methods, +1 (`-reset`).

IdentityCore/tests/MSIDDIContainerTests.m
```diff
@@ -0,0 +1,405 @@
...
+    id<MSIDDIContainerTestProtocol> resolved = [self.container resolveProtocol:@protocol(MSIDDIContainerTestProtocol)];
+
+    XCTAssertEqualObjects([resolved greeting], @"mocked");
+}
+
+- (void)testReset_whenCalled_shouldClearRegistrationsAndCache
+{
+    [self.container registerClass:[MSIDDIContainerTestService class]
+                         lifetime:MSIDDIContainerLifetimeSingleton
+                          factory:^id { return [MSIDDIContainerTestService new]; }];
```
antrix1989, line 190: how will this resolve more complex case, when we need to provide parameters in the init? i.e:

`_authority = [[MSIDAADAuthority alloc] initWithURL:[NSURL URLWithString:authorityString] rawTenant:nil context:nil error:&localError];`

IdentityCore/src/util/MSIDDIContainer.m
```diff
@@ -0,0 +1,296 @@
...
+@property (nonatomic, copy) id _Nonnull (^factory)(void);
+
+@end
+
+@implementation MSIDDIContainerEntry
+@end
+
+@interface MSIDDIContainer ()
+{
+    NSMutableDictionary<NSString *, MSIDDIContainerEntry *> *_entryByKey;
```
antrix1989: can we use properties instead of ivars?

kaisong1990: Switched to `@property` declarations in e55689982 — `nonatomic`, dispatch-queue still does the synchronization. All 21 tests green. 👍

IdentityCore/tests/MSIDDIContainerTests.m
kaisong1990: Good question! The factory block is `id (^)(void)` so it can close over any local state and call any initializer. Two patterns:

**1. Captured-config factory** — register a block that captures the params:
```objc
NSString *authorityString = @"https://login.microsoftonline.com/common";
[container registerClass:[MSIDAADAuthority class]
                lifetime:MSIDDIContainerLifetimeSingleton
                 factory:^id {
    NSError *localError = nil;
    return [[MSIDAADAuthority alloc] initWithURL:[NSURL URLWithString:authorityString]
                                       rawTenant:nil
                                         context:nil
                                           error:&localError];
}];
```

**2. Class-method seam** (`resolveImplClassForClass:`) — when the constructor takes per-call params that vary at the call site, the container just swaps *which class* to instantiate; the caller still owns the `[[Foo alloc] initWith…]` call. This is usually the cleaner fit for the `MSIDAADAuthority` case since the URL/tenant come from request data, not container config.

So the container handles both "fixed deps at registration" (factory) and "variable params at call site" (class seam) without needing a parameterized resolve API.

IdentityCore/src/util/MSIDDIContainer.m
kaisong1990: Also I have update the Forge guidance in this [PR](https://github.com/AzureAD/azure-activedirectory-tokenbroker-for-objc/pull/2403) from broker repo, and it should not use ivar in future. If you are using forge and seeing the ivar patter keeps prompting, please let me know

IdentityCore/tests/MSIDDIContainerTests.m
kaisong1990: Adding a third pattern — **factory-as-dependency** — which directly answers your `MSIDAADAuthority` example where the URL/tenant come from caller data:

```objc
// 1. Define a factory block type
typedef MSIDAADAuthority *(^MSIDAADAuthorityFactory)(NSURL *url, NSString *tenant, NSError **error);

// 2. Production registers the real factory once (e.g., at startup)
[container registerClass:[MSIDAADAuthorityFactory class] // or a dedicated protocol
                lifetime:MSIDDIContainerLifetimeSingleton
                 factory:^id {
    return ^MSIDAADAuthority *(NSURL *url, NSString *tenant, NSError **error) {
        return [[MSIDAADAuthority alloc] initWithURL:url rawTenant:tenant context:nil error:error];
    };
}];

// 3. Caller resolves the factory and invokes it with their own params
MSIDAADAuthorityFactory makeAuthority = [container resolveClass:[MSIDAADAuthorityFactory class]];
NSError *error = nil;
MSIDAADAuthority *authority = makeAuthority([NSURL URLWithString:authorityString], nil, &error);
```

In tests, just register a different factory block that returns a stub/mock authority. This is the standard DI answer (analogous to `Func<T>`/`Provider<T>` injection in other DI containers).

So the full picture for parameterized construction is:
- **Captured-config factory** — params known at registration time
- **Class-method seam** (`resolveImplClassForClass:`) — caller owns `alloc/init`, container picks the class
- **Factory-as-dependency** — caller-provided params at resolve time, container picks the factory

Let me know if you'd like me to add a short "parameterized construction" section to the header doc covering these three patterns.

kaisong1990: Added a **Parameterized construction** section to `MSIDDIContainer.h` covering the three patterns we discussed — captured-config factory, class-method seam (`resolveImplClassForClass:` / `resolveImplClassForProtocol:`), and factory-as-dependency — using `MSIDAADAuthority` as the running example. See commit 3527d7c2c88228f70c39fa63b6befdf65c649b08.

antrix1989 (approved):

PR #1809, opened 2026-05-08
[minor][engg]: Route MSIDThrottlingService through MSIDDIContainer

kaisong1990:
## Summary

Routes the two throttling class-method seams
(`+[MSIDThrottlingService updateLastRefreshTimeDatasource:context:error:]` and
`+[MSIDThrottlingMetaDataCache getLastRefreshTimeWithDatasource:context:error:]`)
through `MSIDDIContainer` so unit tests can inject fakes with
`-registerProtocol:lifetime:factory:` instead of swizzling.

This is the **first concrete adoption** of `MSIDDIContainer` (#1807) and the
reference example for the migration described in:

- Spec: ADO PBI 3571651 (IdentityDivision/Engineering)
- Spec doc: SPEC-MSIDDIContainer-adoption.md §3.3 (`Class<Protocol>` for class-method seams)
- Architecture: ARCH-MSIDDIContainer.md §4 (production never `register*`s; tests only)

## What changed

**Production**
- New protocol `MSIDThrottlingRefreshing` extracted from the static class method
  on `MSIDThrottlingService`. `MSIDThrottlingService` exposes `+resolvedRefresher`
  returning `Class<MSIDThrottlingRefreshing>`, defaulting to `[MSIDThrottlingService class]`.
- New protocol `MSIDThrottlingMetaDataReading` extracted from
  `+getLastRefreshTimeWithDatasource:context:error:` on `MSIDThrottlingMetaDataCache`.
  `MSIDThrottlingMetaDataCache` exposes `+resolvedMetaDataReader` returning
  `Class<MSIDThrottlingMetaDataReading>`, defaulting to itself.
- Refresher call-sites (`MSIDLocalInteractiveController`,
  `MSIDSSOExtensionInteractiveTokenRequestController`,
  `MSIDBrokerInteractiveController`,
  `MSIDXpcInteractiveTokenRequestController`)
  now go through `+resolvedRefresher`.
- The single read call-site in `MSIDThrottlingModelNonRecoverableServerError`
  now goes through `+resolvedMetaDataReader`.

**Tests**
- `MSIDBrokerInteractiveControllerIntegrationTests` registers an inline
  `MSIDFakeThrottlingRefresher` and resets the container in tearDown — replaces
  the prior swizzle.
- `MSIDThrottlingServiceIntegrationTests` (`#if !TARGET_OS_IOS` path) registers
  fakes for both `MSIDThrottlingRefreshing` and `MSIDThrottlingMetaDataReading`
[...]

PR #1838, opened 2026-05-20
[minor][engg]: Harden MSIDDIContainer (thread-safety + protocol conformance check)

kaisong1990:
## Summary

Address Copilot review finding [`#3230317749`](https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1810#discussion_r3230317749) from PR #1810: `MSIDDIContainer.resolveImplClassForProtocol:orDefault:` was returning the resolved `Class` without verifying that it actually conforms to the requested protocol. A test that misregistered a fake class (via the unavoidable `(id)` cast for `Class` values) would silently no-op — the call site received a class that did not implement the expected `+selectors` and the failure surfaced far from the registration mistake.

Tracked by **[AB#3571651](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3571651)** (PBI) / **[AB#3612208](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3612208)** (task). Independent of #1809 and #1810 — `MSIDDIContainer` shipped in #1807 and is the only file touched here.

The companion finding [`#3230317732`](https://github.com/AzureAD/microsoft-authentication-library-common-for-objc/pull/1810#discussion_r3230317732) — singleton-install race — is **already addressed on `dev`**:
- `+sharedInstance` uses `dispatch_once`.
- The per-key singleton install in `-resolveKey:description:defaultProvider:` uses `dispatch_barrier_sync` with double-checked locking (lines ~225–258).

No code change for that half; this PR pins it with a new stress test.

## What changed

`MSIDDIContainer.m`
- `resolveImplClassForProtocol:orDefault:` now calls `[resolvedClass conformsToProtocol:proto]`. On mismatch: `NSAssert` in debug; in release, `MSID_LOG_WITH_CTX(MSIDLogLevelError, …)` and fall back to the caller's `orDefault:` provider so a broken registration cannot brick the production code path.
- `resolveImplClassForClass:orDefault:` deliberately **not** given an analogous `isSubclassOfClass:` check — the class-keyed seam is duck-typed by design (see header) so callers can swap in sibling classes that just share `+selectors`. A subclass-only check would be a public-API regression and was rejected when an existing test (`testResolveImplClassForClass_whenClassFactoryRegistered_shouldReturnRegisteredClass`) exercised exactly that sibling-class pattern.

`MSIDDIContainerTests.m`
- `testResolveImplClassForProtocol_whenRegisteredClassDoesNotConform_shouldAssertInDebug` — negative path, debug build.
- `testResolveImplClassForClass_whenRegisteredClassIsNotSubclass_shouldAssertInDebug` — pins the duck-typed contract: a non-subclass resolves without throwing.
- `testResolveImplClassForProtocol_whenDefaultProviderUsed_shouldNotValidateAgainAfterFallback` — sanity check that the validation targets the registered-factory path, not the default path.
- `testConcurrentRegistrationAndResolution_shouldNotCrashAndConverge` — `dispatch_apply` 100 iterations racing `registerProtocol:` against `resolveImplClassForProtocol:`, then asserts a deterministic final state. Pins the existing `dispatch_barrier_sync` register/reset path.

`changelog.txt` — entry under `TBD`.

## Thread-safety audit (inline, per task acceptance criteria)

| # | Concern | Verdict |
|---|---|---|
| 1 | **Shared mutable state + sync mechanism.** `_entryByKey`, `_singletonCache`, `_synchronizationQueue`. All reads via `dispatch_sync` (concurrent), all writes via `dispatch_barrier_sync` on the same concurrent queue. New conformance check operates on a local `Class` variable captured outside the queue — adds no new shared state. | ✅ Unchanged |
| 2 | **TOCTOU.** Reads of the singleton cache and entry happen inside the same `dispatch_sync` block, so the cache-hit/entry-fetch pair cannot tear. Singleton install re-reads the cache inside the `dispatch_barrier_sync` block (double-checked locking) — no TOCTOU across the read→install boundary. The new `conformsToProtocol:` check runs on the *post*-resolve `Class` value, so even if the registration changes after resolution, we validate exactly what the caller would get. | ✅ |
| 3 | **Re-entrancy / deadlock.** Factory invocations were already moved outside the queue for transient lifetime and inside the barrier (with explicit doc) for singleton lifetime. The new conformance branch invokes `defaultProvider()` on the fallback path **outside** any queue (we've already returned from `-resolveKey:` before reaching the check), so a default provider that re-enters `MSIDDIContainer.sharedInstance` cannot deadlock. | ✅ |
| 4 | **`dispatch_once` token static.** `+sharedInstance` lines 48–56 — `static dispatch_once_t onceToken;`. Unchanged. | ✅ |
| 5 | **`os_unfair_lock` by pointer.** Not used by this container (queue-based). N/A. | ✅ |
| 6 | **Retain cycles in new helper blocks.** No new blocks added in `MSIDDIContainer.m`. The new test fixtures don't capture `self`. | ✅ |
| 7 | **Public API contract unchanged for #1809, #1810, broker.** Method signatures, nullability, return types, and `description` strings unchanged. Behavior change is strictly: a misregistered non-conforming class that *previously* returned and crashed at the call site *now* falls back to `orDefault:` in release / asserts in debug. Existing callers in #1807, #1809, #1810 all register classes that genuinely conform to their protocol keys (verified by the existing test `testResolveImplClassForProtocol_whenClassFactoryRegistered_shouldReturnRegisteredClass` continuing to pass). | ✅ |
| 8 | **Tests pass.** `xcodebuild test -workspace IdentityCore.xcworkspace -scheme "IdentityCore Mac" -only-testing:"IdentityCoreTests Mac/MSIDDIContainerTests"` → **25/25 passed (0 unexpected)** locally on macOS. | ✅ |

## Risk

Low. The validation has an explicit release-mode fallback (`return defaultProvider()`), so even a hypothetically-misregistered production caller degrades gracefully + logs, rather than crashing. The debug assert catches the misregistration in test runs where it matters.

## Out of scope

- Broker submodule pointer bump — none required for this PR.
- `resolveImplClassForClass:` subclass enforcement — explicitly rejected, see body above.

---

Linked: [AB#3571651](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3571651), [AB#3612208](https://identitydivision.visualstudio.com/fac9d424-53d2-45c0-91b5-ef6ba7a6bf26/_workitems/edit/3612208)

### DIFF
--- a/IdentityCore/src/util/MSIDDIContainer.m
+++ b/IdentityCore/src/util/MSIDDIContainer.m
@@ -22,6 +22,7 @@
 // THE SOFTWARE.
 
 #import "MSIDDIContainer.h"
+#import "MSIDLogger+Internal.h"
 
 @interface MSIDDIContainerEntry : NSObject
 
@@ -161,7 +162,26 @@
                        // Cast Class -> id; +resolveKey: validates non-nil.
                        return (id)defaultClass;
                    }];
-    return (Class)resolved;
+
+    Class resolvedClass = (Class)resolved;
+
+    // Validate that the resolved class conforms to the requested protocol.
+    // A misregistered fake class would otherwise return successfully and the
+    // failure would surface as an unrecognized-selector crash at the call
+    // site, far from the registration mistake. Debug: assert so the test
+    // fails loudly. Release: log + fall back to the caller's default so a
+    // broken registration cannot brick the production code path.
+    if (![resolvedClass conformsToProtocol:proto])
+    {
+        NSAssert(NO, @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it",
+                 NSStringFromClass(resolvedClass), NSStringFromProtocol(proto));
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, nil,
+                          @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it; falling back to default",
+                          NSStringFromClass(resolvedClass), NSStringFromProtocol(proto));
+        return defaultProvider();
+    }
+
+    return resolvedClass;
 }
 
 - (Class)resolveImplClassForClass:(Class)cls
@@ -176,6 +196,11 @@
                        Class defaultClass = defaultProvider();
                        return (id)defaultClass;
                    }];
+    // No conformance check here: the class-keyed seam is intentionally
+    // duck-typed (see header) — the registered class need only supply
+    // matching +selectors at runtime, not a subclass relationship. Misuse
+    // surfaces as an unrecognized-selector exception at the call site,
+    // which is the documented contract.
     return (Class)resolved;
 }
 
--- a/IdentityCore/src/util/MSIDDIContainer.m
+++ b/IdentityCore/src/util/MSIDDIContainer.m
@@ -177,6 +177,22 @@
         MSID_LOG_WITH_CTX(MSIDLogLevelError, nil,
                           @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it; falling back to default",
                           NSStringFromClass(resolvedClass), NSStringFromProtocol(proto));
+
+        // Self-heal: evict the bad cached entry under a barrier write so
+        // subsequent resolves return the default cleanly instead of
+        // repeatedly hitting the cached non-conforming class (which would
+        // re-trip the assert in Debug and spam logs + re-call the default
+        // provider in Release). removeObjectForKey: is a safe no-op for
+        // missing keys, so we evict from both maps unconditionally — the
+        // singleton cache is only populated for singleton-lifetime
+        // registrations, but transient registrations simply skip that
+        // dictionary without needing a lifetime branch here.
+        NSString *key = [self keyForProtocol:proto];
+        dispatch_barrier_sync(self.synchronizationQueue, ^{
+            [self.singletonCache removeObjectForKey:key];
+            [self.entryByKey removeObjectForKey:key];
+        });
+
         return defaultProvider();
     }
 
--- a/IdentityCore/tests/MSIDDIContainerTests.m
+++ b/IdentityCore/tests/MSIDDIContainerTests.m
@@ -471,6 +471,12 @@
     // crash or yield a torn class reference.
     NSInteger iterations = 100;
 
+    // XCTest assertions are not guaranteed to be thread-safe, so we collect any
+    // unexpected greetings into a synchronized array from the worker blocks and
+    // assert on the test thread once dispatch_apply has completed.
+    NSMutableArray<NSString *> *unexpectedGreetings = [NSMutableArray array];
+    NSLock *unexpectedGreetingsLock = [NSLock new];
+
     dispatch_apply(iterations, dispatch_get_global_queue(QOS_CLASS_USER_INITIATED, 0), ^(size_t i) {
         if ((i % 2) == 0)
         {
@@ -488,9 +494,18 @@
                                       }];
             // impl is either the registered mock or the default — both conform.
             NSString *greeting = [impl greeting];
-            XCTAssertTrue([greeting isEqualToString:@"mocked-class"] || [greeting isEqualToString:@"hello-class"]);
+            if (![greeting isEqualToString:@"mocked-class"] && ![greeting isEqualToString:@"hello-class"])
+            {
+                [unexpectedGreetingsLock lock];
+                [unexpectedGreetings addObject:greeting ?: @"<nil>"];
+                [unexpectedGreetingsLock unlock];
+            }
         }
     });
+
+    XCTAssertEqual(unexpectedGreetings.count, 0u,
+                   @"Unexpected greetings observed during concurrent resolution: %@",
+                   unexpectedGreetings);
 
     // Deterministic final state: register one last time, resolve, expect mock.
     [self.container registerProtocol:@protocol(MSIDDIContainerTestClassMethodProtocol)
--- a/IdentityCore/tests/MSIDDIContainerTests.m
+++ b/IdentityCore/tests/MSIDDIContainerTests.m
@@ -97,6 +97,15 @@
 
 @end
 
+// Intentionally does NOT conform to MSIDDIContainerTestClassMethodProtocol —
+// used to exercise the conformance-validation path in
+// resolveImplClassForProtocol:orDefault: / resolveImplClassForClass:orDefault:.
+@interface MSIDDIContainerTestNonConformingService : NSObject
+@end
+
+@implementation MSIDDIContainerTestNonConformingService
+@end
+
 #pragma mark - Tests
 
 @interface MSIDDIContainerTests : XCTestCase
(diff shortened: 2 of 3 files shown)


IdentityCore/tests/MSIDDIContainerTests.m
```diff
@@ -402,4 +411,105 @@ - (void)testReset_whenAfterClassFactoryRegistered_shouldFallBackToDefault
...
+                              return [MSIDDIContainerTestClassMethodService class];
+                          }]));
+#else
+    XCTSkip(@"NSAssert is compiled out in release; see fallback-path test");
+#endif
+}
+
+- (void)testResolveImplClassForClass_whenRegisteredClassIsNotSubclass_shouldAssertInDebug
+{
+#if DEBUG
```
kaisong1990: @copilot since UTs is all running under debug macro, please validate m,y state and remove the if else macro if my state is true

IdentityCore/tests/MSIDDIContainerTests.m
```diff
@@ -402,4 +411,105 @@ - (void)testReset_whenAfterClassFactoryRegistered_shouldFallBackToDefault
...
+                             factory:^id { return (id)[MSIDDIContainerTestNonConformingService class]; }];
+
+    XCTAssertThrows(([self.container
+        resolveImplClassForProtocol:@protocol(MSIDDIContainerTestClassMethodProtocol)
+                          orDefault:^Class {
+                              return [MSIDDIContainerTestClassMethodService class];
+                          }]));
+#else
+    XCTSkip(@"NSAssert is compiled out in release; see fallback-path test");
+#endif
```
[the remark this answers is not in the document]
kaisong1990: Updated to remove DEBUG checks as all tests should happen under DEBUG macro

josephpab (commented):
Clean, well-scoped hardening. The conformance check is a good addition — it catches misregistrations at debug time and degrades gracefully in release. Thread-safety audit in the PR body is thorough. Two inline notes below.

IdentityCore/src/util/MSIDDIContainer.m
```diff
@@ -160,7 +161,26 @@ - (Class)resolveImplClassForProtocol:(Protocol *)proto
...
+
+    // Validate that the resolved class conforms to the requested protocol.
+    // A misregistered fake class would otherwise return successfully and the
+    // failure would surface as an unrecognized-selector crash at the call
+    // site, far from the registration mistake. Debug: assert so the test
+    // fails loudly. Release: log + fall back to the caller's default so a
+    // broken registration cannot brick the production code path.
+    if (![resolvedClass conformsToProtocol:proto])
+    {
+        NSAssert(NO, @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it",
```
josephpab, line 175: `Severity: Medium` — **Stale singleton cache entry after conformance failure.**

When a singleton-lifetime factory returns a non-conforming class, the value is already installed in `_singletonCache` by `resolveKey:description:defaultProvider:` before this check runs. Every subsequent `resolveImplClassForProtocol:` call for this key will cache-hit the bad value, fail the conformance check, and re-invoke `defaultProvider()`.

This is unlikely in practice (debug asserts catch it), but in release the log line will fire on every resolve and the default provider is called repeatedly instead of once.

**Recommendation:** Consider clearing the bad entry from the singleton cache on the failure path so subsequent resolves go through the default path cleanly:

```objc
if (![resolvedClass conformsToProtocol:proto])
{
    NSAssert(NO, @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it",
             NSStringFromClass(resolvedClass), NSStringFromProtocol(proto));
    MSID_LOG_WITH_CTX(MSIDLogLevelError, nil,
                      @"MSIDDIContainer: class '%@' resolved for protocol '%@' does not conform to it; falling back to default",
                      NSStringFromClass(resolvedClass), NSStringFromProtocol(proto));
    // Evict the bad entry so subsequent resolves don't repeat this path.
    NSString *key = [self keyForProtocol:proto];
    dispatch_barrier_sync(self.synchronizationQueue, ^{
        [self.entryByKey removeObjectForKey:key];
        [self.singletonCache removeObjectForKey:key];
    });
    return defaultProvider();
}
```

If you intentionally want to preserve the registration (e.g., to keep the log repeating as a signal), a comment documenting that choice would be helpful.

IdentityCore/tests/MSIDDIContainerTests.m
```diff
@@ -402,4 +411,112 @@ - (void)testReset_whenAfterClassFactoryRegistered_shouldFallBackToDefault
...
+    // installed via an (id)-cast. Unit tests run in Debug, so this MUST fire
+    // an assertion when resolved.
+    [self.container registerProtocol:@protocol(MSIDDIContainerTestClassMethodProtocol)
+                            lifetime:MSIDDIContainerLifetimeSingleton
+                             factory:^id { return (id)[MSIDDIContainerTestNonConformingService class]; }];
+
+    XCTAssertThrows(([self.container
+        resolveImplClassForProtocol:@protocol(MSIDDIContainerTestClassMethodProtocol)
+                          orDefault:^Class {
+                              return [MSIDDIContainerTestClassMethodService class];
```
josephpab, line 476: `Severity: Low` — **Consider adding a release-mode fallback test.**

This test pins the debug-assert path (`XCTAssertThrows`), which is valuable. To also pin the release-mode contract (conformance failure → fall back to `defaultProvider` instead of crashing), consider adding a companion test that temporarily installs a custom `NSAssertionHandler` (one that records instead of throwing) and verifies the default class is returned.

This would exercise the log + fallback path that release users actually hit. Not blocking — the current coverage is already solid for the debug build.

josephpab (approved):

IdentityCore/src/util/MSIDDIContainer.m
kaisong1990: Thanks @josephpab — landed in ce8ad4ec7. After NSAssert+log, the failure branch now does a `dispatch_barrier_sync` eviction of `_singletonCache[key]` and `_entryByKey[key]` before returning `defaultProvider()`, so subsequent resolves in Release no longer re-hit the bad cached class (no log spam, deterministic fallback). `removeObjectForKey:` is a safe no-op on missing keys, so no lifetime branching needed.

IdentityCore/tests/MSIDDIContainerTests.m
kaisong1990: Thanks @josephpab — landed in ce8ad4ec7. Added `testResolveImplClassForProtocol_whenRegisteredClassDoesNotConformAndAssertSuppressed_shouldReturnDefaultAndEvictCache`, which swaps in a recording `NSAssertionHandler` via `[[NSThread currentThread] threadDictionary][NSAssertionHandlerKey]` (restored in `@finally`), then asserts: first resolve returns the default, second resolve also returns the default, the handler captured ≥1 failure, and the broken factory was invoked exactly once (pinning the cache eviction from finding 1). The existing `XCTAssertThrows` test is kept as a complementary check.

josephpab (approved):

PR #1849, opened 2026-05-26
[Minor][Feature]: Add support for openid-vc URL handling

josephpab:

- [x] All tests pass locally

## PR Title Format

**Examples:**

## Proposed changes

> **PR scope was revised after initial review.** The original commits introduced `MSIDWebOpenIdVcResponse` + `MSIDWebOpenIdVcResponseOperation` as terminal handlers (mirroring the existing `MSIDWebOpenBrowserResponse` pipeline). That design terminated the auth session with `MSIDErrorSessionCanceledProgrammatically` the moment `openid-vc://` fired, which dismissed the MSAL embedded webview before the wallet round-trip completed — breaking the verifier-driven flow where the same webview is expected to continue and complete the original `acquireToken` call after VID succeeds. This revision keeps the webview alive and replaces the response/operation pipeline with inline handling in `MSIDAADOAuthEmbeddedWebviewController`.

Adds iOS/macOS parity with Android PR [AzureAD/microsoft-authentication-library-common-for-android#3013](https://github.com/AzureAD/microsoft-authentication-library-common-for-android/pull/3013) — fixes [AB#3535641](https://identitydivision.visualstudio.com/Engineering/_workitems/edit/3535641): the AAD embedded webview now intercepts `openid-vc://` (OpenID for Verifiable Credentials) navigations and hands them off to a registered wallet **without dismissing the webview**. The calling app's MSAL `acquireToken` remains a single-shot call from the developer's perspective — the wallet detour is transparent. The verifier's page is expected to drive the webview to its terminal state once the VID exchange resolves, at which point the standard `msauth://` redirect handling completes the request normally.

### How it works

1. **`MSIDAADOAuthEmbeddedWebviewController.decidePolicyAADForNavigationAction:`** detects `openid-vc://` (case-insensitive) and routes to a new private `-handleOpenIdVcNavigationAction:decisionHandler:` helper. The in-webview navigation is cancelled but the webview is left presented and the auth session is not terminated.

2. **Default fallback (non-extension):** the helper mutates the URL by appending Microsoft-namespaced query parameters (`x_ms_caller_redirect_uri`, `x_ms_caller_bundle_id`, `x_ms_correlation_id`) so the wallet can bounce back to the calling app, then calls `[MSIDAppExtensionUtil sharedApplicationOpenURL:]`. Standard OpenID4VP wallets ignore the unknown `x_ms_*` parameters per spec.

3. **Optional delegate (`MSIDOpenIdVcHandling`):** new public protocol on `MSIDAADOAuthEmbeddedWebviewController` (`@property (nonatomic, weak, nullable) id<MSIDOpenIdVcHandling> openIdVcHandler;`) that hosts can implement to take over openid-vc handling entirely. The intended consumer is the Microsoft Authenticator SSO extension when it eventually hosts the VID flow in-process (the extension would present VID UI as a modal on top of the embedded webview — no cross-process round-trip required). When the property is nil (today's only case), behavior is identical to step 2.

4. **App extension path:** terminates the auth session with `MSIDErrorAttemptToOpenURLFromExtension` — extensions can't call `UIApplication.openURL`. If a delegate is attached (future SSO-extension integration), the delegate path runs instead.

### What was removed

The original commit's response/operation pipeline is removed because the new design handles `openid-vc` entirely at the navigation-policy layer — the navigation never produces a webview response object, never reaches the request dispatcher:

- `MSIDWebOpenIdVcResponse.{h,m}`
- `MSIDWebOpenIdVcResponseOperation.{h,m}`
- `MSIDWebOpenIdVcResponseTests.m`
- `MSID_OPEN_OPENID_VC_OPERATION` constant
- `MSIDAADWebviewFactory.m` openid-vc classification branch
- `MSIDInteractiveAuthorizationCodeRequest.m` both V1 and V2 openid-vc dispatcher cases
- `MSIDWebResponseOperationFactoryTests.test_openIdVcResponse_should_return_openIdVcOperation`

### Design decisions

- **Keep MSAL webview alive** during the wallet detour. The verifier's page drives the webview to completion after the VP POSTs to `response_uri`. This makes the wallet detour transparent to MSAL callers.
- **Microsoft-namespaced bring-back parameters** (`x_ms_*`) rather than overloading standard OpenID4VP fields. Non-Microsoft wallets ignore them per spec; the Microsoft wallet reads them to construct the bounce-back URL.
- **Public delegate protocol shipped now for future SSO-extension hosting.** Adding it preemptively means the SSO extension integration can land later by simply setting a property — no further CommonCore changes required. The protocol is additive; default (`nil`) behavior matches today's openURL fallback exactly.
- **No ECS flight** — matches Android PR #3013; can be added later if needed.
- **Embedded webview only** — `ASWebAuthenticationSession` cannot intercept arbitrary `openid-vc://` redirects. Out of scope.
- **Single scheme** — only `openid-vc://`. Other VC schemes (`openid://vc/*`, `vcclient://openid/*`) accepted by Authenticator's `VerifiableCredentialURLHandler` are intentionally deferred.
- **URLs treated as PII** — credential-offer URLs can carry issuer identifiers and configuration parameters. All logging uses `[url msidPIINullifiedURL]`.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

Touches the navigation-policy decision point in the AAD embedded webview, which sits on every interactive MSAL auth flow. Behavior is gated by URL scheme matching: nothing existing changes unless a navigation whose scheme is `openid-vc` is encountered, which doesn't happen today in any AAD flow. The removal of the response/operation pipeline is internal cleanup with no exposed-API impact (the deleted classes were never shipped in a release).

The newly-added `MSIDOpenIdVcHandling` protocol and the `openIdVcHandler` property are additive; consumers that don't opt in (everyone today) get the openURL fallback path, which is functionally a re-implementation of what the original PR's `MSIDWebOpenIdVcResponseOperation` did, minus the `endWebAuthWithURL:` call that was tearing down the webview.

### Test coverage

All in `MSIDAADOAuthEmbeddedWebviewControllerTests` unless noted:

- **URL-mutation behavior (4 new tests):** verifies `x_ms_*` parameters are appended correctly, existing query items are preserved, mutation is idempotent if the params are already present, and the helper no-ops when `endURL` is blank.
- **Decision-handler outcomes (2 existing, refactored):** the two openid-vc navigation tests now use an `MSIDOpenIdVcWebViewControllerSpy` subclass that captures the URL passed to the new private `openOpenIdVcHandoffURL:` method instead of letting `MSIDAppExtensionUtil` crash on a nil `[UIApplication sharedApplication]` in the logic-test runner. They now assert both that the navigation is cancelled AND that the URL was opened.
- **Delegate dispatch (3 new):** verifies the handler is invoked when set (and openURL is NOT called); that an error from the handler completion terminates the auth session via `endWebAuthWithURL:error:`; that a nil handler still runs the existing openURL fallback unchanged.

`MSIDWebOpenIdVcResponseTests` and the `openIdVcResponse_should_return_openIdVcOperation` factory test are removed alongside their subjects. `MSIDAADWebviewFactoryTests` still passes — the openid-vc branch removal there has no behavioral impact (openid-vc URLs no longer become webview responses by design).

### Build validation

- `IdentityCore iOS` scheme builds with `-Werror` ✅
- All 19 `MSIDAADOAuthEmbeddedWebviewControllerTests` (16 existing/preserved + 3 new handler-path tests) pass ✅
- `MSIDAADWebviewFactoryTests` and the trimmed `MSIDWebResponseOperationFactoryTests` pass ✅
- Broker superproject (`ADAuthenticationBroker.xcworkspace`) builds clean against the updated submodule pointer ✅

### Files

**New (1):**
- `IdentityCore/src/webview/MSIDOpenIdVcHandling.h` — `MSIDOpenIdVcHandling` delegate protocol

**Modified (10):**
- `MSIDAADOAuthEmbeddedWebviewController.{h,m}` — openid-vc inline handler + delegate property + nullability annotation
- `MSIDOAuth2EmbeddedWebviewController.{h,m}` — exposed `endURL` as readonly property so the AAD subclass can read the calling app's MSAL redirect URI
- `MSIDWebviewConstants.{h,m}` — added `MSID_SCHEME_OPENID_VC` + three `MSID_OPENID_VC_*` query-parameter keys
- `MSIDAADWebviewFactory.m` — removed openid-vc response classification
- `MSIDInteractiveAuthorizationCodeRequest.m` — removed both openid-vc dispatcher cases
- `MSIDWebResponseOperationConstants.{h,m}` — removed `MSID_OPEN_OPENID_VC_OPERATION`
- `IdentityCore.xcodeproj/project.pbxproj` — added new header, removed deleted source/test files
- `tests/MSIDAADOAuthEmbeddedWebviewControllerTests.m` — added test subclass + 7 new tests
- `tests/integration/ios/MSIDWebResponseOperationFactoryTests.m` — removed openid-vc test
- `changelog.txt` — updated TBD entry

**Deleted (5):**
- `MSIDWebOpenIdVcResponse.{h,m}`
- `MSIDWebOpenIdVcResponseOperation.{h,m}`
- `MSIDWebOpenIdVcResponseTests.m`

Net diff: −153 lines across 18 files.

### Codebase audit

Confirmed no `openid4`-family schemes exist anywhere in the consuming code (`openid4vp://`, `openid4vci://`, `openid-credential-offer://`, `haip://`, `mdoc-openid4vp://`). The "OpenID4VCI" symbols visible in the Microsoft Authenticator's Wallet Library are internal class names, not URL schemes. `openid-vc` remains the only OpenID-for-Verifiable-Credentials URL scheme in use today, matching the Android scope.


josephpab:
@copilot resolve the merge conflicts in this pull request

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
```diff
@@ -234,6 +235,12 @@ - (MSIDWebviewResponse *)oAuthResponseWithURL:(NSURL *)url
     MSIDWebWPJResponse *wpjResponse = [[MSIDWebWPJResponse alloc] initWithURL:url context:context error:nil];
     if (wpjResponse) return wpjResponse;
     
+    // Try to create an OpenID Verifiable Credentials (openid-vc://) response
+    MSIDWebOpenIdVcResponse *openIdVcResponse = [[MSIDWebOpenIdVcResponse alloc] initWithURL:url
+                                                                                     context:context
+                                                                                       error:nil];
```
antonioalwan: The error is not captured, it is preferred to use the &error pattern and check `if (openIDVcResponse && !error) else {log(error details)}` even though existing code doesn't do that.

changelog.txt
```diff
@@ -14,6 +14,7 @@ Version 1.24.0
 * Close cookies popup in automation tests #1814
 * Exclude swift files from macOS targets #1827
 * Propagate client_data from broker error response for legacy broker support (#1834)
+* Handle openid-vc:// URLs in the AAD embedded webview by handing them off to the system (#1849)
```
antonioalwan: Should this line go into TBD section

galwarsamy (commented):
Nice work — clean interception design, solid test coverage. Couple of things below.

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.m
```diff
@@ -307,6 +314,145 @@ - (NSURL *)urlBySwappingScheme:(NSURL *)url
...
+                        completion:^(NSError * _Nullable error)
+        {
+            if (error)
+            {
+                MSID_LOG_WITH_CTX_PII(MSIDLogLevelWarning, self.context,
+                                      @"openid-vc handler reported error; ending auth session: %@",
+                                      MSID_PII_LOG_MASKABLE(error));
+                [self endWebAuthWithURL:nil error:error];
+            }
+        }];
```
galwarsamy, line 365: Strong `self` capture inside the async completion block here. The handler property is `weak` so the classic retain cycle won't form, but the block still extends the controller's lifetime for as long as the handler holds it — which could be a long time once the SSO extension hosts VID in-process.

Suggest the usual `__weak`/`__strong` dance:

```objc
__weak typeof(self) weakSelf = self;
[handler handleOpenIdVcURL:requestURL
         webviewController:self
         callerRedirectUri:self.endURL.absoluteString
             correlationId:self.context.correlationId
                completion:^(NSError * _Nullable error)
{
    __strong typeof(weakSelf) strongSelf = weakSelf;
    if (!strongSelf) return;

    if (error)
    {
        MSID_LOG_WITH_CTX_PII(MSIDLogLevelWarning, strongSelf.context,
                              @"openid-vc handler reported error; ending auth session: %@",
                              MSID_PII_LOG_MASKABLE(error));
        [strongSelf endWebAuthWithURL:nil error:error];
    }
}];
```

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.m
```diff
@@ -307,6 +314,145 @@ - (NSURL *)urlBySwappingScheme:(NSURL *)url
...
+    // Cancel the in-webview navigation but DO NOT call completeWebAuthWithURL: or
+    // endWebAuthWithURL:. The webview stays presented so the user returns to it
+    // after the wallet completes. The verifier's page is responsible for driving
+    // the webview to its terminal state once the VID exchange succeeds.
+    decisionHandler(WKNavigationActionPolicyCancel);
+#else
+    MSID_LOG_WITH_CTX(MSIDLogLevelWarning, self.context,
+                      @"openid-vc:// scheme is not supported on this platform");
+    decisionHandler(WKNavigationActionPolicyCancel);
+#endif
```
galwarsamy, line 405: On macOS without a handler, this just logs + cancels the navigation but leaves the auth session alive and the webview presented. The verifier page will sit there waiting for the wallet round-trip that never happened — effectively a silent hang from the user's perspective.

Should we terminate the session with an error here so MSAL surfaces something actionable to the caller? Something like:

```objc
#else
    NSError *unsupportedError = MSIDCreateError(MSIDErrorDomain,
                                                MSIDErrorInternal,
                                                @"openid-vc:// handling requires a registered openIdVcHandler on this platform",
                                                nil, nil, nil, self.context.correlationId, nil, YES);
    [self endWebAuthWithURL:nil error:unsupportedError];
    decisionHandler(WKNavigationActionPolicyCancel);
#endif
```

Or if this is intentionally deferred until macOS VID ships, a `// TODO:` with tracking context would help.

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.h
```diff
@@ -26,16 +26,29 @@
...
+/// webview. When set, the controller delegates handoff entirely to the handler
+/// (allowing in-process VID UI to be presented on top of the webview without
+/// any cross-process round trip). When nil, the controller falls back to its
+/// default behavior of mutating the URL with Microsoft-namespaced query
+/// parameters and dispatching `UIApplication.openURL` to a registered wallet.
+@property (nonatomic, weak, nullable) id<MSIDOpenIdVcHandling> openIdVcHandler;
+
 - (BOOL)decidePolicyAADForNavigationAction:(WKNavigationAction *)navigationAction
-                           decisionHandler:(void (^)(WKNavigationActionPolicy))decisionHandler;
+                           decisionHandler:(nullable void (^)(WKNavigationActionPolicy))decisionHandler;
```
galwarsamy, line 48: Nit (non-blocking): `decisionHandler` is now `nullable` here, but `handleOpenIdVcNavigationAction:decisionHandler:` calls it unconditionally in 3 places. Won't matter in practice (WKWebView always passes non-nil) but if we're advertising nullable in the public API, might be worth a nil guard in the private helper for defensive consistency.

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.m
josephpab: macOS is out of scope for VID flows by product decision — no openid-vc:// is expected to reach this branch on Mac. Keeping the current log + cancel; the comment now says so explicitly. Happy to revisit if VID later expands to Mac.

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.h
josephpab: Done — added `if (decisionHandler)` guards around all three call sites in `handleOpenIdVcNavigationAction:decisionHandler:` (handler path, extension path, openURL fallback, and the no-op-on-Mac branch). Matches the nullable annotation in the public header.

IdentityCore/src/webview/embeddedWebview/MSIDAADOAuthEmbeddedWebviewController.m
josephpab: Done in the next push. Added the `__weak` / `__strong` dance with an early-return on nil `strongSelf`, matching the pattern you suggested.

IdentityCore/src/webview/operations/MSIDWebOpenIdVcResponseOperation.m
```diff
@@ -0,0 +1,94 @@
...
+        }
+        return YES;
+    }
+
+    MSID_LOG_WITH_CTX_PII(MSIDLogLevelInfo, nil, @"Opening openid-vc URL - %@", [self.openIdVcURL msidPIINullifiedURL]);
+    [MSIDAppExtensionUtil sharedApplicationOpenURL:self.openIdVcURL];
+
+    if (error)
+    {
+        *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorSessionCanceledProgrammatically, @"Authorization session was cancelled programatically.", nil, nil, nil, correlationId, nil, YES);
```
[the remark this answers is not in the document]
josephpab: Stale after the architectural revision: `MSIDWebOpenIdVcResponseOperation.{h,m}` is deleted in this PR's current state. The new design handles openid-vc inline in `MSIDAADOAuthEmbeddedWebviewController` and never produces a response object. See the updated PR description for the rationale.

IdentityCore/tests/integration/ios/MSIDWebResponseOperationFactoryTests.m
```diff
@@ -59,4 +61,27 @@ - (void)test_openBroswerResponse_should_return_openBroserOperation
...
+
+    XCTAssertNotNil(webResponse);
+    XCTAssertTrue([webResponse isKindOfClass:MSIDWebOpenIdVcResponse.class]);
+
+    MSIDWebResponseBaseOperation *operation = [MSIDWebResponseOperationFactory createOperationForResponse:webResponse
+                                                                                                    error:&error];
+    XCTAssertNil(error);
+    XCTAssertNotNil(operation);
+    XCTAssertTrue([operation isKindOfClass:MSIDWebOpenIdVcResponseOperation.class]);
+    [MSIDWebResponseOperationFactory unRegisterforResponse:webResponse];
```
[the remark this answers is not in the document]
josephpab: Stale after the architectural revision: `test_openIdVcResponse_should_return_openIdVcOperation` is deleted, since `MSIDWebOpenIdVcResponse` itself no longer exists. The other test in the same file (`test_openBroswerResponse_should_return_openBroserOperation`) still does the same `unRegisterforResponse:` cleanup; this comment's point would still apply there if it ever matters, but no changes from me on that.

IdentityCore/src/oauth2/aad_base/MSIDAADWebviewFactory.m
josephpab: Stale after the architectural revision: the entire openid-vc response-classification branch is removed from `MSIDAADWebviewFactory.responseWithURL:` (commit 079e82b86). The new design handles openid-vc upstream in `MSIDAADOAuthEmbeddedWebviewController` so the factory never sees the URL.

changelog.txt
josephpab: The changelog entry was already moved to (and rewritten in) the `TBD` section as part of the architectural revision — see commit 079e82b86. Closing out.

galwarsamy (approved):

PR #1862, opened 2026-06-12
Mobile Onboarding Non Broker Enrollment

swasti29:
Mobile Onboarding Non Broker Enrollment

### DIFF
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.h
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.h
@@ -25,11 +25,12 @@
 #import "MSIDBaseRequestController.h"
 #import "MSIDTokenRequestProviding.h"
 #import "MSIDRequestControlling.h"
+#import "MSIDWebviewNavigationDelegate.h"
 
 @class MSIDInteractiveTokenRequestParameters;
 @class MSIDWebWPJResponse;
 
-@interface MSIDLocalInteractiveController : MSIDBaseRequestController <MSIDRequestControlling>
+@interface MSIDLocalInteractiveController : MSIDBaseRequestController <MSIDRequestControlling, MSIDWebviewNavigationDelegate>
 
 @property (nonatomic, readonly, nullable) MSIDInteractiveTokenRequestParameters *interactiveRequestParamaters;
 
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
@@ -318,6 +318,17 @@
     // Could not build a controller, cannot retry the request.
     if (!brokerController)
     {
+        if (!brokerError)
+        {
+            brokerError = MSIDCreateError(MSIDErrorDomain,
+                                          MSIDErrorInternal,
+                                          @"Failed to resolve a controller after MDM enrollment.",
+                                          nil, nil, nil,
+                                          self.requestParameters.correlationId,
+                                          nil,
+                                          YES);
+        }
+
         MSID_LOG_WITH_CTX(MSIDLogLevelError, self.requestParameters,
                           @"Failed to resolve a controller after MDM enrollment: %@", brokerError);
         CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], brokerError);
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
@@ -291,12 +291,13 @@
         MSID_LOG_WITH_CTX(MSIDLogLevelError, self.requestParameters,
                           @"MDM enrollment failed (status=%@).", status);
 
+        NSString *errorDescription = [NSString stringWithFormat:@"MDM enrollment failed with status: %@", status];
         NSError *enrollmentError = MSIDCreateError(MSIDErrorDomain,
                                                    MSIDErrorInternal,
-                                                   @"MDM enrollment failed",
+                                                   errorDescription,
                                                    nil, nil, nil,
                                                    self.requestParameters.correlationId,
-                                                   @{ @"mdm_enrollment_status": status },
+                                                   nil,
                                                    NO);
 
         CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], enrollmentError);
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
@@ -71,7 +71,10 @@
         __weak typeof(self) weakSelf = self;
         parameters.webviewConfigurationBlock = ^(id<MSIDWebviewInteracting> webviewController) {
             __strong typeof(self) strongSelf = weakSelf;
-            if (!strongSelf) return;
+            if (!strongSelf)
+            {
+                return;
+            }
 
             [strongSelf.navigationHandler configureWebviewController:webviewController
                                                           delegate:strongSelf];
--- a/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
+++ b/IdentityCore/src/controllers/MSIDLocalInteractiveController.m
@@ -37,11 +37,15 @@
 #import "MSIDWebWPJResponse.h"
 #import "MSIDWebUpgradeRegResponse.h"
 #import "MSIDThrottlingService.h"
+#import "MSIDWebviewNavigationHandler.h"
+#import "MSIDWebMDMEnrollmentCompletionResponse.h"
+#import "MSIDRequestControllerFactory.h"
 
 @interface MSIDLocalInteractiveController()
 
 @property (nonatomic, readwrite) MSIDInteractiveTokenRequestParameters *interactiveRequestParamaters;
 @property (nonatomic) MSIDInteractiveTokenRequest *currentRequest;
+@property (nonatomic, strong) MSIDWebviewNavigationHandler *navigationHandler;
 
 @end
 
@@ -61,6 +65,17 @@
     if (self)
     {
         _interactiveRequestParamaters = parameters;
+        _navigationHandler = [[MSIDWebviewNavigationHandler alloc] initWithContext:parameters];
+
+        // Wire the navigation handler into the webview controller once it's created.
+        __weak typeof(self) weakSelf = self;
+        parameters.webviewConfigurationBlock = ^(id<MSIDWebviewInteracting> webviewController) {
+            __strong typeof(self) strongSelf = weakSelf;
+            if (!strongSelf) return;
+
+            [strongSelf.navigationHandler configureWebviewController:webviewController
+                                                          delegate:strongSelf];
+        };
     }
 
     return self;
@@ -225,11 +240,24 @@
     
     [request executeRequestWithCompletion:^(MSIDTokenResult *result, NSError *error, MSIDWebviewResponse *msauthResponse)
     {
-        if (msauthResponse && [msauthResponse isKindOfClass:[MSIDWebWPJResponse class]])
+        if (msauthResponse)
         {
             self.currentRequest = nil;
-            [self handleWebMSAuthResponse:(MSIDWebWPJResponse *)msauthResponse completion:completionBlock];
-            return;
+            
+            // Handle MDM enrollment completion response
+            if ([msauthResponse isKindOfClass:MSIDWebMDMEnrollmentCompletionResponse.class])
+            {
+                [self handleWebMDMEnrollmentCompletionResponse:(MSIDWebMDMEnrollmentCompletionResponse *)msauthResponse
+                                                    completion:completionBlock];
+                return;
+            }
+            
+            if ([msauthResponse isKindOfClass:[MSIDWebWPJResponse class]])
+            {
+                [self handleWebMSAuthResponse:(MSIDWebWPJResponse *)msauthResponse
+                                   completion:completionBlock];
+                return;
+            }
         }
         
 #if !EXCLUDE_FROM_MSALCPP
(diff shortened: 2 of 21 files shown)


IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -61,6 +65,17 @@ - (nullable instancetype)initWithInteractiveRequestParameters:(nonnull MSIDInter
     if (self)
     {
         _interactiveRequestParamaters = parameters;
+        _navigationHandler = [[MSIDWebviewNavigationHandler alloc] initWithContext:parameters];
+
+        // Wire the navigation handler into the webview controller once it's created.
+        __weak typeof(self) weakSelf = self;
+        parameters.webviewConfigurationBlock = ^(id<MSIDWebviewInteracting> webviewController) {
```
josephpab, line 72: **Severity: High – `webviewConfigurationBlock` is unconditionally overwritten**

**Issue:** The init assigns `parameters.webviewConfigurationBlock = ^...` without checking whether the caller has already supplied a block.

**Impact:** `webviewConfigurationBlock` is documented as a public extension point in `MSIDInteractiveTokenRequestParameters.h:64–69` ("Optional configuration block called when webview controller is created. Use this to set delegates, customize behavior, etc."). Any caller (a future MSAL public surface, an internal test fixture, or a sibling controller layer) that sets a block on `parameters` before constructing this controller has their block silently discarded — the only consumer is `MSIDInteractiveAuthorizationCodeRequest.m:147`, which invokes whatever block is set at request time, so the loss is invisible until somebody loses a feature.

**Recommendation:** Compose with any existing block instead of replacing it:
```objc
MSIDWebviewConfigurationBlock existingBlock = [parameters.webviewConfigurationBlock copy];
__weak typeof(self) weakSelf = self;
parameters.webviewConfigurationBlock = ^(id<MSIDWebviewInteracting> webviewController) {
    __strong typeof(self) strongSelf = weakSelf;
    if (strongSelf)
    {
        [strongSelf.navigationHandler configureWebviewController:webviewController
                                                        delegate:strongSelf];
    }
    if (existingBlock) existingBlock(webviewController);
};
```
Add a unit test that sets a block on `parameters` before constructing the controller and asserts both blocks run.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -243,4 +271,103 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
+
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, self.requestParameters,
+                          @"Failed to resolve a controller after MDM enrollment: %@", brokerError);
+        CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], brokerError);
+        completionBlock(nil, brokerError);
+        return;
+    }
+
+    // Stop local telemetry, the downstream controller owns its own event.
+    CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], nil);
```
josephpab, line 344: **Severity: High – Telemetry reports success before the retry runs**

**Issue:** `CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], nil)` stops the local controller's event with a `nil` error (success) before `[brokerController acquireToken:completionBlock]` has even started.

**Impact:** If the broker controller's retry fails, telemetry shows the local controller "succeeded" while the user-facing request failed. Two events fire for one logical `acquireToken` call with the same `telemetryApiId`, and the first one always reports success regardless of the real outcome. This breaks success-rate metrics and obscures MDM-enrollment-then-retry failures in dashboards.

**Recommendation:** Pick one of:
1. **Preferred:** Wrap `completionBlock` so the local event is stopped with the *actual* retry outcome:
   ```objc
   [brokerController acquireToken:^(MSIDTokenResult *result, NSError *retryError) {
       CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], retryError);
       completionBlock(result, retryError);
   }];
   ```
2. Follow the precedent in `promptBrokerInstallWithResponse:` (lines ~196–201, same file): just hand off to the broker controller without stopping the local event, and let the downstream event represent the request.

Picking one consistent pattern across both handoffs (WPJ + MDM) avoids the "first event always succeeds" wart.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -243,4 +271,103 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
+    }
+
+    // MDM enrollment complete. Retry the token request through the appropriate controller.
+    // If broker is installed and SSO extension is active, the factory returns the SSO controller.
+    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters,
+                      @"MDM enrollment complete (status=%@); retrying token request.", status);
+
+    NSError *brokerError = nil;
+    id<MSIDRequestControlling> brokerController =
+        [MSIDRequestControllerFactory interactiveControllerForParameters:self.interactiveRequestParamaters
```
josephpab, line 318: **Severity: Medium – No retry-limit / loop guard for MDM completion**

**Issue:** `interactiveControllerForParameters:` returns either a broker controller or another `MSIDLocalInteractiveController` (see `MSIDRequestControllerFactory.m:311–331`). On platforms without broker, when `shouldUseBroker == NO`, or when the SSO extension isn't active, the retry produces a fresh `MSIDLocalInteractiveController`, which re-runs the same `init` (re-overwriting `webviewConfigurationBlock`) and re-registers itself as the navigation delegate.

**Impact:** If the eSTS / Intune flow returns the MDM completion URL a second time (server bug, eventually-consistent enrollment state, misconfigured tenant), the controller drives the webview again, gets the same response, and recurses — no structural stop. The user sees the webview re-open with no way out and the completion never fires.

**Recommendation:** Add a single-shot guard. E.g., add a `BOOL mdmEnrollmentRetryAttempted` on `MSIDInteractiveTokenRequestParameters`; check it here before delegating to the factory; set it to YES before `[brokerController acquireToken:]`. On a second MDM completion, short-circuit to the failure path:
```objc
if (self.interactiveRequestParamaters.mdmEnrollmentRetryAttempted)
{
    // Already retried once — server keeps returning the enrollment URL, treat as failure.
    ...completionBlock(nil, retriedTwiceError);
    return;
}
self.interactiveRequestParamaters.mdmEnrollmentRetryAttempted = YES;
```
Add a regression test that returns the MDM completion response twice and asserts the second one errors out without re-driving the factory.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -243,4 +271,103 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
+    NSString *status = mdmEnrollmentCompletionResponse.status ?: @"<none>";
+
+    // Failure path: MDM enrollment did not complete.
+    if (!mdmEnrollmentCompletionResponse.isSuccess)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, self.requestParameters,
+                          @"MDM enrollment failed (status=%@).", status);
+
+        NSError *enrollmentError = MSIDCreateError(MSIDErrorDomain,
+                                                   MSIDErrorInternal,
```
josephpab, line 299: **Severity: Medium – MDM enrollment failure is reported as generic `MSIDErrorInternal`**

**Issue:** `MDM enrollment failed` and `Failed to resolve a controller after MDM enrollment` (line 324) both use `MSIDErrorInternal`, the same code used for unrelated internal failures across IdentityCore (allocation failures, programmer errors, etc.).

**Impact:** Callers (MSAL public layer, customer apps that adopt MSAL) cannot programmatically distinguish "user did not complete MDM enrollment" from any other internal failure without string-matching the description or sniffing the `@"mdm_enrollment_status"` userInfo key. They lose the ability to surface a meaningful UX (e.g., "please complete enrollment in the Settings app") or to choose whether to retry interactively.

**Recommendation:** Introduce a dedicated error code (e.g., `MSIDErrorMDMEnrollmentFailed`) in `MSIDError.h` and use it here. This mirrors how the sibling `handleWebMSAuthResponse:` already promotes distinct failures (`MSIDErrorWorkplaceJoinRequired`, `MSIDErrorInsufficientDeviceStrength`) to first-class codes. Keep the human-readable status in userInfo for diagnostics, but make the failure mode itself branch-able.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -243,4 +271,103 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
+                                                   @"MDM enrollment failed",
+                                                   nil, nil, nil,
+                                                   self.requestParameters.correlationId,
+                                                   @{ @"mdm_enrollment_status": status },
```
josephpab: **Severity: Medium – Bare string literal for userInfo key**

**Issue:** `@"mdm_enrollment_status"` is a bare string literal here, duplicated in `MSIDInteractiveControllerIntegrationTests.m:980` and `:1026`.

**Impact:** No build-time protection against typos in either production or tests. Any future consumer (MSAL public layer, broker handoff code, customer app) must rediscover the key from the source. Diverges from the existing IdentityCore convention of named constants for `NSError.userInfo` keys (`MSIDUserDisplayableIdkey`, `MSIDHomeAccountIdkey`, `MSIDTokenProtectionRequired`, etc.).

**Recommendation:** Promote to a named constant — natural home is `MSIDWebMDMEnrollmentCompletionResponse.h` (it owns the semantic) or `MSIDWebviewConstants.h`:
```objc
extern NSString * const MSIDMDMEnrollmentStatusUserInfoKey;
```
Reference it from both the controller and the tests so the two stay in sync.

josephpab (commented):
Thanks for the PR! A few items worth attention before merging — mostly around the lifecycle/correctness of the new MDM completion handler and one spot where the init silently overwrites a documented extension point. Summary:

- **High:** The local controller's `init` clobbers `parameters.webviewConfigurationBlock`, which is documented as an extension point for callers.
- **High:** Local telemetry is stopped with `nil` (success) before the broker controller's retry actually runs — misclassifies failed retries.
- **Medium:** No retry-limit / single-shot guard on the MDM completion → completion → MDM completion loop. The factory can hand back another `MSIDLocalInteractiveController` on platforms without broker.
- **Medium:** MDM enrollment failure is bucketed under generic `MSIDErrorInternal`; callers can't branch on it.
- **Medium:** `@"mdm_enrollment_status"` is a bare string literal duplicated across production code and tests; promote to a named constant.

Nothing blocking the design — the wire / class-prefix / Allman-brace conventions all look good, and the test coverage for the new handler is genuinely thorough (factory-nil, retry-fails, missing-status, etc.). See inline comments for specifics.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
swasti29: It is always sets new if sets earlier ignore by design and also this is set in constructor

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
swasti29: This telemetry is not for this purpose it is for response overall.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
swasti29: This already handle in MDM Response class

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
swasti29: Updated the error and the unit tests

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
swasti29: Updated the error and the unit tests, no longer needed userinfo unnecessary complicated things

IdentityCore/tests/integration/MSIDInteractiveControllerIntegrationTests.m
```diff
@@ -556,4 +563,439 @@ - (void)testAcquireToken_whenInvalidBrokerInstallRequest_shouldReturnError
...
+    [MSIDTestSwizzle classMethod:@selector(interactiveControllerForParameters:tokenRequestProvider:error:)
+                           class:[MSIDRequestControllerFactory class]
+                           block:(id)^(__unused id obj,
+                                       __unused MSIDInteractiveTokenRequestParameters *p,
+                                       __unused id<MSIDTokenRequestProviding> tp,
+                                       NSError **err)
+    {
+        factoryCallCount++;
+        if (err) *err = factoryError;
+        return nil;
```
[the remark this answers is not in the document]
swasti29: Fixed in 3aabef7.

IdentityCore/src/controllers/MSIDLocalInteractiveController.h
```diff
@@ -25,11 +25,12 @@
 #import "MSIDBaseRequestController.h"
```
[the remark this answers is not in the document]
swasti29: Changelog will be updated as part of the release process.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -61,6 +65,17 @@ - (nullable instancetype)initWithInteractiveRequestParameters:(nonnull MSIDInter
...
+
+        // Wire the navigation handler into the webview controller once it's created.
+        __weak typeof(self) weakSelf = self;
+        parameters.webviewConfigurationBlock = ^(id<MSIDWebviewInteracting> webviewController) {
+            __strong typeof(self) strongSelf = weakSelf;
+            if (!strongSelf) return;
+
+            [strongSelf.navigationHandler configureWebviewController:webviewController
+                                                          delegate:strongSelf];
+        };
```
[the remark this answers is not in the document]
swasti29: This is intentional. The block is set here during initialization and this is the only place it's assigned — callers don't set it externally for this controller type. Fixed brace style in 3aabef7.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -225,11 +240,24 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
             self.currentRequest = nil;
-            [self handleWebMSAuthResponse:(MSIDWebWPJResponse *)msauthResponse completion:completionBlock];
-            return;
+            
+            // Handle MDM enrollment completion response
+            if ([msauthResponse isKindOfClass:MSIDWebMDMEnrollmentCompletionResponse.class])
+            {
+                [self handleWebMDMEnrollmentCompletionResponse:(MSIDWebMDMEnrollmentCompletionResponse *)msauthResponse
+                                                    completion:completionBlock];
+                return;
```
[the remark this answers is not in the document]
swasti29: Tests already exist: MSIDMDMEnrollmentCompletionResponseTests.m covers response parsing, and MSIDInteractiveControllerIntegrationTests.m:801 covers the handler (testHandleWebMDMEnrollmentCompletionResponse_whenCompletionBlockIsNil_shouldReturnSafely).

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
```diff
@@ -243,4 +271,92 @@ - (void)acquireTokenWithRequest:(MSIDInteractiveTokenRequest *)request
...
+
+    // Could not build a controller, cannot retry the request.
+    if (!brokerController)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, self.requestParameters,
+                          @"Failed to resolve a controller after MDM enrollment: %@", brokerError);
+        CONDITIONAL_STOP_TELEMETRY_EVENT([self telemetryAPIEvent], brokerError);
+        completionBlock(nil, brokerError);
+        return;
+    }
```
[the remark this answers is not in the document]
swasti29: Already handled — the code checks if (!brokerError) and creates a fallback error via MSIDCreateError before calling the completion block. No nil error is passed to callers.

IdentityCore/src/controllers/MSIDLocalInteractiveController.m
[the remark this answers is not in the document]
swasti29: Intentional — this controller owns the webviewConfigurationBlock assignment. No external caller sets it before this point. Fixed brace style in 3aabef7.

josephpab (approved):

PR #1875, opened 2026-06-24
[minor] [feature]: Introduce silent routing + shared Browser Core Components

agubuzomaximus:

- [x] All tests pass locally

## Proposed changes
part of POC:

this PR introduces imporved test & logging for MSIDBoundTokenProvider. Silent flow has been implemented while interactive flow is sutbbed. An upcoming PR will implement the interactive flow. 

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [X] Small – No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)


IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -0,0 +1,493 @@
...
+        return nil;
+    }
+
+    MSIDAuthenticationScheme *authScheme = request.authScheme ?: [MSIDAuthenticationScheme new];
+    NSOrderedSet<NSString *> *scopes = [request.scopes msidScopeSet];
+    NSUUID *correlationId = context.correlationId ?: [NSUUID UUID];
+
+    NSError *parametersError = nil;
+    // TODO: split OIDC scopes (openid/profile/offline_access) from resource scopes once the
+    // interactive broker-flip path is wired; the silent redemption path is unaffected.
```
hieunguyenmsft: curious why splitting need, IIRC broker should already have generic scope handling.

agubuzomaximus: Thanks Hieu — I am reusing the shared bucket/union/validator handling, not duplicating it; the split is only needed here because the browser-native GetToken request arrives as one flat `scopes` string with no oidcScope field, so MSAL's `defaultOIDCScopes` injection never runs on this path. It is a no-op for the silent flow (hence `oidcScopes:nil` today) and only gets wired when the interactive broker-flip lands, where I will source the default set from a shared constant so we do not re-declare the list.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -26,7 +26,32 @@
...
+#import "MSIDTokenResult.h"
+#import "MSIDAccessToken.h"
+#import "MSIDAccount.h"
+#import "MSIDTokenResponse.h"
+#import "MSIDBrowserNativeMessageGetTokenResponse.h"
+#import "MSIDBrokerOperationTokenResponse.h"
+
+#if TARGET_OS_IPHONE
+#import "MSIDKeychainTokenCache.h"
+#import "MSIDLegacyTokenCacheAccessor.h"
```
ameyapat: Is this required?

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -37,25 +62,44 @@ - (void)acquireBoundTokenWithRequest:(MSIDBrowserNativeMessageGetTokenRequest *)
                      completionBlock:(MSIDBoundTokenProviderCompletionBlock)completionBlock
 {
     NSParameterAssert(completionBlock);
-    if (!completionBlock) return;
-
+    if (!completionBlock)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"%@ completionBlock is nil; cannot deliver result.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
```
ameyapat: nit: Add a helper private method to log and call it here. In case, log format needs to change, it will allow change only in one place.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+                                                      telemetryApiId:nil
+                                                       brokerOptions:nil
+                                                         requestType:MSIDRequestBrokeredType
+                                                 intuneAppIdentifier:nil
+                                                               error:&parametersError];
+    if (!parameters)
+    {
+        if (error)
+        {
+            *error = parametersError;
```
ameyapat: Use MSIDCreateError to create a new error and pass parametersError as subError

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+        return NO;
+    }
+
+    if (!parameters.accountIdentifier)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context, @"%@ No account identifier on request; silent path not possible.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
+        return NO;
+    }
+
+    return [self hasCachedTokenForParameters:parameters context:context];
```
ameyapat: If you are querying keychain for a token, it is an expensive operation. Here we are not using the queried token and it will be inefficient

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+                                   request:(MSIDBrowserNativeMessageGetTokenRequest *)request
+                                   context:(nullable id<MSIDRequestContext>)context
+                           completionBlock:(MSIDBoundTokenProviderCompletionBlock)completionBlock
+{
+    MSIDDefaultTokenCacheAccessor *tokenCache = [self defaultTokenCache:context];
+    MSIDAccountMetadataCacheAccessor *accountMetadataCache = [self accountMetadataCache:context];
+    if (!tokenCache || !accountMetadataCache)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, context, @"%@ Token cache unavailable; routing to interactive path.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
+        [self acquireTokenInteractivelyWithParameters:parameters request:request context:context completionBlock:completionBlock];
```
ameyapat: If silent fails we should throw interaction_required error and have calling trigger interactive request and not trigger it automatically ourselves

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+            {
+                completionBlock(nil, shapeError);
+            }
+            return;
+        }
+
+        if ([error.domain isEqualToString:MSIDErrorDomain] && error.code == MSIDErrorInteractionRequired)
+        {
+            MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context, @"%@ Silent path requires user interaction; falling back to interactive path.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
+            [strongSelf acquireTokenInteractivelyWithParameters:parameters request:request context:context completionBlock:completionBlock];
```
ameyapat: Same here, we shouldn't fallback to interactive but return interaction required to caller

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+                                     @"Failed to build GetToken response payload.",
+                                     nil, nil, nil, result.correlationId, nil, NO);
+        }
+        return nil;
     }
 
     NSError *serializationError = nil;
-    NSData *data = [NSJSONSerialization dataWithJSONObject:payload options:0 error:&serializationError];
-    if (serializationError || !data)
+    NSData *data = [NSJSONSerialization dataWithJSONObject:responseDictionary options:0 error:&serializationError];
```
ameyapat: Call https://developer.apple.com/documentation/foundation/jsonserialization/isvalidjsonobject(_:)?language=objc before attempting to convert to JSON

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+        if (error)
+        {
+            *error = serializationError ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
+                                                           @"Failed to serialize GetToken response payload.",
+                                                           nil, nil, nil, result.correlationId, nil, NO);
+        }
+        return nil;
     }
 
     return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
```
ameyapat: Does the returned serialized string need to have "  escaped to \"

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+                                             request:(MSIDBrowserNativeMessageGetTokenRequest *)request
+{
+    MSIDAccessToken *accessToken = result.accessToken;
+    if (!accessToken)
+    {
+        return nil;
+    }
+
+    NSMutableDictionary *response = [NSMutableDictionary new];
+    response[@"access_token"] = accessToken.accessToken ?: @"";
```
ameyapat: Instead of empty string, we shouldn't populate access_token or the other missing fields at all so that required checks fail later.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+    account[@"userName"] = result.account.username ?: request.loginHint;
+    response[@"account"] = account;
+    response[@"state"] = request.state ?: @"";
+
+    return response;
+}
+
+#pragma mark - Dependencies
+
+// Builds the default token cache accessor backed by the com.microsoft.adalcache keychain group.
```
ameyapat: Can you confirm if com.microsoft.adalcache is being used? I'm not sure if such a group exists or is used

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+    }
+
+    // getRefreshTokenWithAccount: prefers a Bound App Refresh Token when one is cached. Only a bound
+    // token makes silent redemption viable here; a regular refresh token is intentionally rejected.
+    MSIDRefreshToken *refreshToken = [tokenCache getRefreshTokenWithAccount:parameters.accountIdentifier
+                                                                  familyId:nil
+                                                             configuration:parameters.msidConfiguration
+                                                                   context:context
+                                                                     error:&cacheError];
+    if ([refreshToken isKindOfClass:[MSIDBoundRefreshToken class]])
```
antrix1989: can we avoid casting? We can add isBound property to the MSIDRefreshToken class

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +130,363 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+        getTokenResponse.state = request.state;
+        getTokenResponse.requestAccountUpn = result.account.username ?: request.loginHint;
+        return [getTokenResponse jsonDictionary];
+    }
+
+    // Cache hit (no fresh server response): shape the payload from the cached result directly.
+    return [self responseDictionaryFromCachedResult:result request:request];
+}
+
+- (NSDictionary *)responseDictionaryFromCachedResult:(MSIDTokenResult *)result
```
antrix1989: can we move it out to separate response object class?

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: so in MSIDKeychainTokenCache the code sets the default KeyChainGroup for ios to 
-> NSString *const MSIDAdalKeychainGroup = @"com.microsoft.adalcache";

and I use the two imports for initalizing the data source then also initalizing the keychain accessor.
so set the data source to the default keychain group then initalize keychain with that datasource. 
so its should be required for the logic here since this is where we interacti with our BART keys.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: I do not think the log format will change. all logs are following MSID_LOG_WITH_CTX unless i am misunderstanding your ask.

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDInteractiveTokenRequestParameters+BrowserNativeMessageGetToken.m
```diff
@@ -0,0 +1,140 @@
...
+    NSMutableDictionary *extraParameters = [request.extraParameters mutableCopy];
+    NSString *childClientId = extraParameters[MSIDBrowserNativeMessageChildClientIdKey];
+    NSString *childRedirectUri = extraParameters[MSIDBrowserNativeMessageChildRedirectUriKey];
+    NSString *nestedAuthBrokerClientId = nil;
+    NSString *nestedAuthBrokerRedirectUri = nil;
+
+    if (![NSString msidIsStringNilOrBlank:childClientId]
+        && ![NSString msidIsStringNilOrBlank:childRedirectUri])
+    {
+        nestedAuthBrokerClientId = clientId;
```
antrix1989, line 87: Keep original comments from the broker, i.e: " Map child_* params to brk_* params:" etc

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDBrowserNativeMessageGetTokenRequestParametersFactory.h
```diff
@@ -0,0 +1,46 @@
...
+#import "MSIDConstants.h"
+
+@class MSIDBrowserNativeMessageGetTokenRequest;
+@class MSIDInteractiveTokenRequestParameters;
+
+NS_ASSUME_NONNULL_BEGIN
+
+@interface MSIDBrowserNativeMessageGetTokenRequestParametersFactory : NSObject
+
++ (nullable MSIDInteractiveTokenRequestParameters *)requestParametersWithRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
```
antrix1989: don't use static methods, use singleton + class methods instead. This way we can mock the class without using swizzling

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDBrowserNativeMessageGetTokenRoutingPolicy.h
```diff
@@ -0,0 +1,61 @@
...
+{
+    MSIDBrowserNativeMessageGetTokenRouteSilent = 0,
+    MSIDBrowserNativeMessageGetTokenRouteInteractive,
+    MSIDBrowserNativeMessageGetTokenRouteUIBlocked,
+    MSIDBrowserNativeMessageGetTokenRouteInteractionRequired
+};
+
+@interface MSIDBrowserNativeMessageGetTokenRoutingPolicy : NSObject
+
++ (MSIDBrowserNativeMessageGetTokenRoute)routeWithForceInteractive:(BOOL)forceInteractive
```
antrix1989: same -- don't use static methods

IdentityCore/src/broker_operation/response/browser_native_message_response/MSIDBrowserNativeMessageGetTokenResponse.h
```diff
@@ -34,6 +34,9 @@ NS_ASSUME_NONNULL_BEGIN
 
 - (instancetype)initWithDeviceInfo:(nullable MSIDDeviceInfo *)deviceInfo NS_UNAVAILABLE;
 - (instancetype _Nullable)initWithTokenResponse:(nonnull MSIDBrokerOperationTokenResponse *)tokenResponse;
+- (instancetype _Nullable)initWithTokenResponse:(nonnull MSIDBrokerOperationTokenResponse *)tokenResponse
```
antrix1989: why do we need this new initializer?

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +183,284 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+    NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
+                                     @"Interactive broker flip is required but not yet implemented in MSIDBoundTokenProvider.",
+                                     nil, nil, nil, context.correlationId, nil, NO);
+    completionBlock(nil, error);
+}
+
+#pragma mark - Response shaping
+
+// Serializes a token result into the browser-native-message GetToken response payload (JSON string).
+- (NSString *)responsePayloadFromResult:(MSIDTokenResult *)result
```
antrix1989: Why don't we use MSIDBrowserNativeMessageGetTokenResponse class and keep parsing there?

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
antrix1989: We can acquire "non" bound Token for SPA apps as well, right? Why do we call it bound?

antrix1989:
I suggest to implement MSIDSPATokenProvider class and move out all logic related to mapping SPA requests/responses there. This class should be 80% identical to ADBBrowserNativeMessageGetTokenOperation, but without broker related classes. The remaining 20% will be ADBSilentTokenOperation/ADBInteractiveTokenOperation +HUP logic in case of broker and MSIDInteractiveTokenRequest/MSIDSilentTokenRequest in case of the local app flow:

Broker/SSO ext path:
SPA Request -> Broker -> Common Core (SPATokenProvider) -> Broker Acquire Token Flow (silent/interactive)

Local App flow (no broker)
SPA Request -> Broker -> Common Core (SPATokenProvider) -> Local Acquire Token Flow (silent/interactive)

swasti29:
@agubuzomaximus Please close this if not relevant anymore

swasti29:
@agubuzomaximus Please close this if not relevant anymore

agubuzomaximus:
> @agubuzomaximus Please close this if not relevant anymore

@swasti29  still in progress. I had to pause due to picking up Juans Mobile Onboarding work. I'll aim to close this week.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: This flow is meant to support Bound App Refresh Tokens so only PRT v4 minted tokens are allowed.

IdentityCore/src/broker_operation/response/browser_native_message_response/MSIDBrowserNativeMessageGetTokenResponse.m
```diff
@@ -54,6 +61,41 @@ - (instancetype)initWithTokenResponse:(MSIDBrokerOperationTokenResponse *)operat
...
+    }
+
+    return self;
+}
+
+- (instancetype)initWithCachedTokenResult:(MSIDTokenResult *)tokenResult
+                                    state:(NSString *)state
+                fallbackRequestAccountUpn:(NSString *)fallbackRequestAccountUpn
+{
+    if (!tokenResult.accessToken)
```
antrix1989: We have "none" AT in "Lookup"mode. Also is it required according to the protocol? Could it be that we just return ID Token?

IdentityCore/src/broker_operation/response/browser_native_message_response/MSIDBrowserNativeMessageGetTokenResponse.m
```diff
@@ -54,6 +61,41 @@ - (instancetype)initWithTokenResponse:(MSIDBrokerOperationTokenResponse *)operat
     return self;
 }
 
+- (instancetype)initWithTokenResponse:(MSIDBrokerOperationTokenResponse *)operationTokenResponse
```
antrix1989: can we define designated initializer?

IdentityCore/src/broker_operation/response/browser_native_message_response/MSIDBrowserNativeMessageGetTokenResponse.m
```diff
@@ -89,4 +137,74 @@ - (NSDictionary *)jsonDictionary
     return response;
 }
 
+// Shapes the GetToken payload from a cached token result (access token cache hit). Optional fields
+// are omitted when blank/nil so downstream required-field validation can fail cleanly rather than
+// receiving empty placeholder values.
+- (NSDictionary *)cachedResultJsonDictionary
```
antrix1989: Currently we use MSIDTokenResult & MSIDBrokerOperationTokenResponse. Can we unify and have dependency only on 1 type of the response in this class? For example MSIDTokenResult. And create a mapping from MSIDBrokerOperationTokenResponse to MSIDTokenResult (we can use category MSIDTokenResult+MSIDBrokerOperationTokenResponse). Another option -- define all properties in this class and use them for json serialization/deserialization.

IdentityCore/src/oauth2/token/MSIDSPATokenAcquiring.h
```diff
@@ -0,0 +1,117 @@
...
+
+/// Normalized acquisition outcome produced by an @c MSIDSPATokenAcquiring backend and consumed by
+/// @c MSIDSPATokenProvider, which feeds it to @c MSIDBrowserNativeMessageGetTokenResponse so response
+/// shaping stays single-sourced.
+///
+/// A backend sets exactly one of:
+///  - @c tokenResult            (local app flow: the provider maps the fresh-vs-cached shape exactly
+///                               as the previous @c MSIDBoundTokenProvider did), or
+///  - @c operationTokenResponse (broker flow: a freshly redeemed broker operation token response).
+@interface MSIDSPATokenAcquisitionResult : NSObject
```
antrix1989, line 42: Can we just use MSIDTokenResult? We should be able to create MSIDTokenResult from MSIDBrokerOperationTokenResponse

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDBrowserNativeMessageGetTokenRequestParametersFactory.m
```diff
@@ -0,0 +1,159 @@
...
+#import "MSIDRequestParameters.h"
+#import "MSIDOAuth2Constants.h"
+#import "NSOrderedSet+MSIDExtensions.h"
+#import "NSString+MSIDExtensions.h"
+#import "MSIDError.h"
+
+static NSString * const MSIDBrowserNativeMessageChildClientIdKey = @"child_client_id";
+static NSString * const MSIDBrowserNativeMessageChildRedirectUriKey = @"child_redirect_uri";
+
+@implementation MSIDBrowserNativeMessageGetTokenRequestParametersFactory
```
antrix1989: can we use category of MSIDInteractiveTokenRequestParameters instead for mapping? See MSIDRequestParameters (ADBroker) as example

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDBrowserNativeMessageGetTokenRoutingPolicy.m
```diff
@@ -0,0 +1,116 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+//
+//------------------------------------------------------------------------------
+
+#import "MSIDBrowserNativeMessageGetTokenRoutingPolicy.h"
+#import "MSIDAccountIdentifier.h"
+
+@implementation MSIDBrowserNativeMessageGetTokenRoutingPolicy
```
antrix1989, line 32: I understand the intention of this class, but I don't think it cover the logic that we have in broker. In broker we have this method:
`- (ADBOperation *)createTokenOperation:(BOOL)isSts canShowUI:(BOOL)canShowUI needsUi:(BOOL *)needsUi requestParameters:(MSIDInteractiveTokenRequestParameters **)requestParams error:(NSError **)error`

It does the following:

1. Create correct operation( request): Silent or Interactive
2. Tells if the operation will need UI. This logic is quite complex, we cannot tell just based on the provided params whether we need UI or not. For example: based on the provided params it seems that we just need silent operation, but if we actually invoke that silent operation and it return interaction_required error then this value should be set to YES.
3. We return error if we cannot create the operation (based on the provided params).

Lets sync offline on this

antrix1989:
Let's also create a broker PR to use this. It will help us catch any logic we might have missed. Also for singletons, lets use MSIDDIContainer and manage singleton there:

1. Remove all "sharedInstance" from the singleton class
2. add this to register it as a singleton:
@implementation MySingletonClass
 
```
+ (void)load
{
    MSIDDIContainer *container = [MSIDDIContainer sharedInstance];
    
    [container registerClass:[MySingletonClass class]
                    lifetime:MSIDDIContainerLifetimeSingleton
                     factory:^id { return [[MySingletonClass alloc] init]; }];
}
```
 
 
then in all other classes when you need to use it, call this:
 
```
MSIDDIContainer *container = [MSIDDIContainer sharedInstance];
self.singletonClass = [container resolveClass:[MySingletonClass class]]; 

```

IdentityCore/src/broker_operation/request/browser_native_message_request/MSIDBrowserNativeMessageGetTokenRoutingPolicy.m
antrix1989: Discussed offline, proposed changes:
1. Keep this class (nit: maybe find better name since we don't use "Routing" term in our code. Instead we use "Request/Flow")
2. Keep only 1st and 2nd values (silent/interactive). All other should be returned as NSErrors.

agubuzomaximus: MSIDBrowserNativeMessageGetTokenRouteSilent = route silent
MSIDBrowserNativeMessageGetTokenRouteInteractive = route interactive

MSIDBrowserNativeMessageGetTokenRouteInteractionRequired = User interaction is required, but showing UI is not allowed ... and this logs with an error when used. MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired, @"User interaction is required.", nil, nil, nil, self.request.correlationId, nil, NO);

MSIDBrowserNativeMessageGetTokenRouteUIBlocked = Shared routing policy resolved to UI-blocked; interaction required but canShowUI is NO. ADBBrowserNativeMessageErrorResponseStatusUiNotAllowed

agubuzomaximus:
here is the broker PR as draft till this is merged in: https://github.com/AzureAD/azure-activedirectory-tokenbroker-for-objc/pull/2581/changes

agubuzomaximus:
I'll close this PR and split the work into smaller bits and chunks.

PR #1883, opened 2026-07-02
[patch][bugfix]: Validate cloud_instance_host_name against known Microsoft identity host list

kaisong1990:
### Summary

`cloud_instance_host_name` from the auth webview redirect URL was previously used to build the cloud authority without checking whether it was a recognized Microsoft identity host. This PR makes that handling more accurate: the value is only used when it is a known AAD public/sovereign cloud host, or a network environment already discovered via instance metadata.

### Changes
- Add a recognized-host check in `setCloudAuthorityWithCloudHostName:` (`MSIDRequestParameters`) and `authorityWithUpdatedCloudHostInstanceName:error:` (`MSIDAADAuthority`). Unrecognized values are ignored and logged; the originally configured authority continues to be used.
- Host matching is case-insensitive to align with cached `preferred_network` values (which may not be case-normalized).
- Add unit tests for known public-cloud, known sovereign-cloud, unknown, nil, blank, and different-case host inputs.
- Update the instance-aware interactive token request integration test to use a valid sovereign-cloud host.

### Impact
- Sovereign-cloud and instance-aware flows are unaffected.
- No public API changes.

### DIFF
--- a/IdentityCore/src/parameters/MSIDRequestParameters.m
+++ b/IdentityCore/src/parameters/MSIDRequestParameters.m
@@ -34,6 +34,8 @@
 #import "MSIDAccountIdentifier.h"
 #import "MSIDIntuneApplicationStateManager.h"
 #import "MSIDAuthenticationScheme.h"
+#import "MSIDAADNetworkConfiguration.h"
+#import "MSIDAadAuthorityCache.h"
 
 @implementation MSIDRequestParameters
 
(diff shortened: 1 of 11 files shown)


IdentityCore/tests/MSIDAADAuthorityTests.m
```diff
@@ -779,6 +779,74 @@ - (void)testNeedsUpdateToHomeAuthority_consumers_shouldReturnTrue
...
+    MSIDAADAuthority *authority = (MSIDAADAuthority *)[@"https://login.microsoftonline.com/common" aadAuthority];
+    NSError *error = nil;
+
+    MSIDAuthority *updatedAuthority = [authority authorityWithUpdatedCloudHostInstanceName:@"login.contoso-sovereign.com" error:&error];
+
+    XCTAssertNotNil(updatedAuthority);
+    XCTAssertNil(error);
+    XCTAssertEqualObjects(updatedAuthority.environment, @"login.contoso-sovereign.com");
+
+    cache.allCloudNetworkEnvironments = savedEnvironments;
```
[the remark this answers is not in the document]
kaisong1990: Good catch — fixed in 0c7ca187c. Wrapped the cache mutation and assertions in `@try/@finally` so `MSIDAadAuthorityCache.sharedInstance` is always restored even if an assertion raises, so the global state can't leak into other tests.

IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -172,13 +173,25 @@ - (NSURL *)tokenEndpoint
 - (void)setCloudAuthorityWithCloudHostName:(NSString *)cloudHostName
 {
     if ([NSString msidIsStringNilOrBlank:cloudHostName]) return;
+
+    NSString *lowercaseHostName = cloudHostName.lowercaseString;
+
+    // Only rewrite the cloud authority when cloud_instance_host_name is a recognized
+    // Microsoft identity host; unrecognized values are ignored so the originally
+    // configured authority continues to be used.
+    if (![MSIDAADAuthority isRecognizedMicrosoftIdentityHost:lowercaseHostName])
```
antrix1989: This will work only for AAD authorities. Lets make isRecognizedMicrosoftIdentityHost object method and implement in all Authority types (similar to "authorityWithUpdatedCloudHostInstanceName" method)

kaisong1990: Good call — done in d6f9dd305. `isRecognizedMicrosoftIdentityHost:` is now an instance method on `MSIDAuthority` (declared in `MSIDAuthority+Internal.h`), mirroring `authorityWithUpdatedCloudHostInstanceName:`: the base implementation returns `NO`, and `MSIDAADAuthority` overrides it with the known-host allow-list check. The call sites now dispatch on the authority instance (`self.authority` here, `self` in the AAD override), so behavior is driven by the authority type rather than hardcoded to AAD. Removed the now-unused `MSIDAADAuthority.h` import, and added a non-AAD (ADFS) test to cover the base default.

IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -172,13 +172,25 @@ - (NSURL *)tokenEndpoint
...
     if ([NSString msidIsStringNilOrBlank:cloudHostName]) return;
+
+    NSString *lowercaseHostName = cloudHostName.lowercaseString;
+
+    // Only rewrite the cloud authority when cloud_instance_host_name is a recognized
+    // Microsoft identity host; unrecognized values are ignored so the originally
+    // configured authority continues to be used.
+    if (![self.authority isRecognizedMicrosoftIdentityHost:lowercaseHostName])
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, nil, @"Ignoring cloud_instance_host_name: host is not a recognized Microsoft identity host.");
```
agubuzomaximus: Thanks for the change Kai. One thing: when we ignore a host we log with a `nil` context, so if a real sovereign cloud ever gets bounced here we've got no correlation id to trace it by. `MSIDRequestParameters` already conforms to `MSIDRequestContext`, so passing `self` (and maybe an execution-flow tag like the PkeyAuth change does) would make these silent rejections easy to spot in the field.

kaisong1990: Thanks @agubuzomaximus — good catch. Fixed in a2fcc5eb1:

- The "Ignoring cloud_instance_host_name" warning now passes `self` as the context instead of `nil`, so it carries the correlation id (`MSIDRequestParameters` conforms to `MSIDRequestContext`). I also switched the sibling authority-creation error log to `self` for consistency.
- Added an execution-flow tag exactly like the PkeyAuth change: a new `MSIDCloudInstanceHostNameIgnoredTag` recorded via `MSIDExecutionFlowInsertTag` when a correlation id is available, so ignored hosts are easy to spot in the field. The 5-char code is left as `UNTAGGED` for the tagging GitHub Action to assign.
- Added a unit test asserting the tag is recorded on the ignored-host path.

antrix1989 (approved):

PR #1887, opened 2026-07-08
[patch][diagnostics]: Add MSIDXpcCanPerformFailureReason diagnostics and complete customized XPC transport failures with errors

kaisong1990:
## Summary

Fixes ADO 3236668: https://dev.azure.com/IdentityDivision/_workitems/edit/3236668

`+[MSIDXpcSingleSignOnProvider canPerformRequest:]` has 5 distinct exit points that all collapse into a bare `NO`, making it impossible to root-cause the production spike of this client-side XPC gate rejecting requests without further instrumentation. This PR adds an additive, non-breaking `reason:` overload that reports which branch fired — at both the provider level and the controller level, since production callers (e.g. OneAuth's `MSAIBrokerClient`) call the controller's bare `canPerformRequest` class method directly and never see CommonCore's internal `MSID_LOG_WITH_CTX` output.

## Changes

- **New enum `MSIDXpcCanPerformFailureReason`** (`MSIDXpcSingleSignOnProvider.h`), with one case per failure branch:
  - `None` (success)
  - `NoProviderInstalled` — no MacBrokerApp/CompanyPortal XPC component on device
  - `DeviceInfoRequestCreationFailed` — SSOExtension `getDeviceInfo` request object failed to construct
  - `DeviceInfoHandshakeError` — handshake completed with a hard error
  - `DeviceInfoHandshakeTimeout` — handshake did not complete within the 1s `dispatch_group_wait` window
  - `ValidateCacheProviderFailed` — no installed provider matches cached/available config
  - `UnsupportedOSVersion` — controller-level gate rejects the request on macOS < 13, before ever reaching the provider
- **New `MSIDXpcCanPerformFailureReasonToString(reason)`** helper for non-PII logging.
- **New `+[MSIDXpcSingleSignOnProvider canPerformRequest:reason:]`** overload that sets `*reason` at each return site. The original no-`reason` `canPerformRequest:` selector is kept as a thin wrapper (`reason:nil`) — fully source/behavior compatible with existing call sites and tests.
- **Split timeout vs. hard-error**: both previously fell through to the same generic `validateCacheXpcProvider` failure; they're now tracked separately and reported as distinct reasons (timeout takes priority if both occurred).
- **Updated both callers** — `MSIDXpcSilentTokenRequestController` and `MSIDXpcInteractiveTokenRequestController` `canPerformRequest` — to call the new overload and log the reason (name + int, no PII) via `MSID_LOG_WITH_CTX` when the result is `NO`.
- **New `+[MSIDXpcSilentTokenRequestController canPerformRequest:]`** and **`+[MSIDXpcInteractiveTokenRequestController canPerformRequest:]`** overloads (reason out-param), mirroring the provider-level pattern, so external callers outside CommonCore (e.g. OneAuth) can retrieve and surface the failure reason to their own telemetry instead of only getting a bare `BOOL`. The existing no-arg `canPerformRequest` selectors remain thin wrappers (`canPerformRequest:nil`) — fully source/behavior compatible.
- **Unit tests**: added coverage in `MSIDXpcSingleSignOnProviderTest.m` for all 5 provider-level failure branches and the success path.
- **Changelog**: added/updated an entry under `TBD`.

## Testing

Ran `./build.py --targets mac_library --no-xcpretty --no-clean`:
- Build succeeded.
- All 2496 tests passed (0 failures).

## Notes

- No PII is logged — only the reason enum name/int and class/method name.
- No behavior change for existing callers of any no-`reason` selector (provider or controller level).
- Adoption in OneAuth (`MSAIBrokerClient.m`) to actually consume the new controller-level `reason:` overload and feed it into OneAuth's own telemetry (`MSAIBrokerEligibilityResponseFactory` / `appendExecutionFlow:`) is tracked separately, since that's a different repo/consumer.


IdentityCore/src/util/mac/MSIDXpcSingleSignOnProvider.m
```diff
@@ -67,6 +67,22 @@
 #import "MSIDXpcProviderCaching.h"
 #import "MSIDFlightManager.h"
 
+static const NSTimeInterval MSIDXpcDispatcherEndpointLookupTimeout = 10.0;
+static const NSTimeInterval MSIDXpcBrokerReplyTimeout = 60.0;
```
ameyapat, line 71: MSIDXpcBrokerReplyTimeout = 60.0 is scheduled around handleXpcWithRequestParams: (line 328), and this same handleRequestParam: path is used by the interactive request (MSIDSSOXpcInteractiveTokenRequest.performXpcRequest → handleRequestParam:

For interactive auth the broker reply is gated on real user interaction (password, MFA, consent), so a hard 60s cap can abort a legitimate in-progress flow and surface a spurious MSIDErrorBrokerXpcUnexpectedError. Is this intended? 

If not, please exempt the interactive path (or use a much longer/disabled timeout there) and add a test asserting an interactive request is not cancelled by this timeout.

ameyapat (approved):

IdentityCore/src/util/mac/MSIDXpcSingleSignOnProvider.m
kaisong1990: Good catch — fixed in 4f121ced63df822e1e55000f5bc372830b6ef931. Interactive broker requests now pass `MSIDXpcBrokerReplyTimeoutDisabled` (0.0) down a shared private funnel into `attemptBrokerRequest:`, which skips scheduling the 60s broker-reply watchdog entirely, so a legitimate in-progress interactive flow (password/MFA/consent) can't be aborted with a spurious `MSIDErrorBrokerXpcUnexpectedError`. Silent requests keep the 60s watchdog unchanged. A connection interruption/invalidation still completes the interactive request via the existing transport-failure path. Added `testInteractiveRequest_doesNotScheduleBrokerReplyWatchdog_delayedReplyStillCompletes` (asserts no 60s schedule + a delayed reply still completes) and `testSilentRequest_schedulesBrokerReplyWatchdog` as a regression guard. Public .h signatures unchanged.

ameyapat (approved):

PR #1894, opened 2026-07-10
[patch] [bugfix]: Fix MSIDFlightManager reader data race on flightProvider

josephpab:

- [x] All tests pass locally

## Proposed changes

`MSIDFlightManager`'s `boolForKey:` and `stringForKey:` tested `self.flightProvider` through the (unsynchronized) property getter **before** dispatching onto the synchronization queue, then re-read it inside the queue:

```objc
if (self.flightProvider) {                      // unsynchronized read
    dispatch_sync(self.synchronizationQueue, ^{
        result = [self.flightProvider ...];     // second read
    });
}
```

That first read races the barrier write in `setFlightProvider:`. When another thread swaps/releases the provider concurrently with a read, the read can end up messaging a freed provider (use-after-free).

In the Microsoft Authenticator app this surfaced as an `objc_msgSend` crash inside `-[MSIDAccountCredentialCache checkFRTEnabled:error:]` at the flight-status read (`stringForKey:MSID_FLIGHT_CLIENT_SFRT_STATUS`), driven from a launch-time background thread that reads FRT/flight state while the main flow assigns a new `flightProvider` from a parsed broker device-info response.

This change reads `_flightProvider` **once from inside the synchronization queue** into a strong local and messages that local, removing the unsynchronized access and keeping the provider alive for the duration of the call. It complements #1876 (which made the setter synchronous via `dispatch_barrier_sync`); the setter and readers are now fully serialized through the queue.

Adds a concurrency regression test (`testConcurrentReads_whileProviderSwappedAndCleared_doNotCrash`) that interleaves provider swaps (including `nil`) with concurrent `boolForKey:`/`stringForKey:` reads.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

All 15 `MSIDFlightManagerTests` pass locally (iOS, Xcode 26.5, iPhone 17 simulator), including the new regression test. Changelog updated under `TBD`.

### DIFF
--- a/IdentityCore/src/MSIDFlightManager.m
+++ b/IdentityCore/src/MSIDFlightManager.m
@@ -120,7 +120,7 @@
 
 #pragma mark - MSIDFlightManagerInterface
 
-- (BOOL)boolForKey:(nonnull NSString *)flightKey 
+- (BOOL)boolForKey:(nonnull NSString *)flightKey
 {
     __block BOOL result = NO;
     // Read the provider only from within the synchronization queue. Testing it via the
--- a/IdentityCore/src/MSIDFlightManager.m
+++ b/IdentityCore/src/MSIDFlightManager.m
@@ -34,6 +34,8 @@
 
 @implementation MSIDFlightManager
 
+@synthesize flightProvider = _flightProvider;
+
 + (instancetype)sharedInstance
 {
     static MSIDFlightManager *sharedInstance = nil;
@@ -116,6 +118,19 @@
     dispatch_barrier_sync(self.synchronizationQueue, ^{
         self->_flightProvider = flightProvider;
     });
+}
+
+- (id<MSIDFlightManagerInterface>)flightProvider
+{
+    __block id<MSIDFlightManagerInterface> flightProvider = nil;
+    // Read on the synchronization queue so external callers are serialized against the
+    // barrier write in setFlightProvider:, matching the internal readers below. Capturing
+    // into a strong local keeps the provider alive for the duration of the call.
+    dispatch_sync(self.synchronizationQueue, ^{
+        flightProvider = self->_flightProvider;
+    });
+    
+    return flightProvider;
 }
 
 #pragma mark - MSIDFlightManagerInterface
--- a/IdentityCore/src/MSIDFlightManager.m
+++ b/IdentityCore/src/MSIDFlightManager.m
@@ -123,25 +123,33 @@
 - (BOOL)boolForKey:(nonnull NSString *)flightKey 
 {
     __block BOOL result = NO;
-    if (self.flightProvider)
-    {
-        dispatch_sync(self.synchronizationQueue, ^{
-            result = [self.flightProvider boolForKey:flightKey];
-        });
-    }
+    // Read the provider only from within the synchronization queue. Testing it via the
+    // unsynchronized property getter first would race the barrier write in setFlightProvider:,
+    // and capturing it into a strong local keeps it alive for the duration of the call.
+    dispatch_sync(self.synchronizationQueue, ^{
+        id<MSIDFlightManagerInterface> flightProvider = self->_flightProvider;
+        if (flightProvider)
+        {
+            result = [flightProvider boolForKey:flightKey];
+        }
+    });
     
     return result;
 }
 
 - (nullable NSString *)stringForKey:(nonnull NSString *)flightKey
 {
-    __block NSString* result = nil;
-    if (self.flightProvider)
-    {
-        dispatch_sync(self.synchronizationQueue, ^{
-            result = [self.flightProvider stringForKey:flightKey];
-        });
-    }
+    __block NSString *result = nil;
+    // Read the provider only from within the synchronization queue. Testing it via the
+    // unsynchronized property getter first would race the barrier write in setFlightProvider:,
+    // and capturing it into a strong local keeps it alive for the duration of the call.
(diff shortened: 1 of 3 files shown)


IdentityCore/tests/MSIDFlightManagerTests.swift
```diff
@@ -208,6 +208,44 @@ class MSIDFlightManagerTests: XCTestCase {
...
+        let expectation = XCTestExpectation(description: "All operations complete")
+        expectation.expectedFulfillmentCount = 300
+        
+        // Interleave provider swaps (including nil) with concurrent reads. Before the reader
+        // fix, boolForKey:/stringForKey: tested the provider through the unsynchronized getter
+        // before dispatching onto the synchronization queue, so a swap-and-release racing a
+        // read could leave the read messaging a freed provider. Reading the provider once from
+        // inside the queue into a strong local removes that window.
+        for i in 0..<100 {
+            DispatchQueue.global().async {
```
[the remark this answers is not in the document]
josephpab: Done in 1dd0b74ff — the count is now derived: `expectedFulfillmentCount = iterations * operationsPerIteration`, so it stays in sync if the loop trip count changes.

IdentityCore/src/MSIDFlightManager.m
```diff
@@ -123,25 +123,33 @@ - (void)setFlightProvider:(id<MSIDFlightManagerInterface>)flightProvider
 - (BOOL)boolForKey:(nonnull NSString *)flightKey
```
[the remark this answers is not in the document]
josephpab: Done in 1dd0b74ff — removed the trailing whitespace on the `boolForKey:` declaration line.

IdentityCore/tests/MSIDFlightManagerTests.swift
```diff
@@ -208,6 +208,46 @@ class MSIDFlightManagerTests: XCTestCase {
...
+                let provider = MockFlightProvider()
+                provider.identifier = "provider-\(i)"
+                provider.boolValues["bool-key"] = true
+                provider.stringValues["string-key"] = "value-\(i)"
+                flightManager.flightProvider = (i % 2 == 0) ? provider : nil
+                expectation.fulfill()
+            }
+            
+            DispatchQueue.global().async {
+                _ = flightManager.bool(forKey: "bool-key")
```
kaisong1990, line 235: This stress test concurrently mutates unsynchronized mock properties, which introduces a test-side data race and can fail under Thread Sanitizer. Please use a read-only or synchronized provider stub.

IdentityCore/src/MSIDFlightManager.m
```diff
@@ -120,28 +120,36 @@ - (void)setFlightProvider:(id<MSIDFlightManagerInterface>)flightProvider
...
-    {
-        dispatch_sync(self.synchronizationQueue, ^{
-            result = [self.flightProvider boolForKey:flightKey];
-        });
-    }
+    // Read the provider only from within the synchronization queue. Testing it via the
+    // unsynchronized property getter first would race the barrier write in setFlightProvider:,
+    // and capturing it into a strong local keeps it alive for the duration of the call.
+    dispatch_sync(self.synchronizationQueue, ^{
+        id<MSIDFlightManagerInterface> flightProvider = self->_flightProvider;
```
kaisong1990, line 145: This is the right fix — reading `_flightProvider` inside the queue serializes it against the barrier setter, and the strong local keeps it alive for the whole call. One thing to keep in mind: the public `flightProvider` getter is still unsynchronized, so external callers could theoretically still race it. Out of scope here though. 👍

kaisong1990 (approved):
Approved with minor suggestions

IdentityCore/tests/MSIDFlightManagerTests.swift
josephpab: Fixed in 91a2a8a74 by taking the *synchronized provider stub* option: `MockFlightProvider`'s call-tracking state (`_boolForKeyCalled`/`_stringForKeyCalled`/`_lastBoolKey`/`_lastStringKey`) is now guarded by an `NSLock`, so concurrent reads through `bool(forKey:)`/`string(forKey:)` no longer race on it. `boolValues`/`stringValues` are only written single-threaded before the provider is published, then read concurrently (concurrent reads are race-free), so they stay lock-free.

Verified the whole `MSIDFlightManagerTests` suite (all 15 cases) now passes under Thread Sanitizer with zero data races (iPhone 16, iOS 18.5, `-enableThreadSanitizer YES`). This also cleared the same TSan races TSan surfaced in the pre-existing `testConcurrentFlightOperations_AreThreadSafe` and `testFlightProviderSetter_IsThreadSafe`, which shared the identical root cause.

IdentityCore/src/MSIDFlightManager.m
josephpab: Went ahead and closed the gap in 91a2a8a74: the public `flightProvider` getter is now overridden to read `_flightProvider` on the synchronization queue (`dispatch_sync`), mirroring the internal readers, so external callers are serialized against the `dispatch_barrier_sync` in `setFlightProvider:` too. Added `@synthesize flightProvider = _flightProvider;` since both accessors are now user-implemented (autosynthesis no longer applies). Changelog entry updated to mention the getter.

kaisong1990 (approved):

PR #1895, opened 2026-07-10
[patch] [bugfix]: Harden msidIsStringNilOrBlank against non-string input

josephpab:

- [x] All tests pass locally

## Proposed changes

`+[NSString(MSIDExtensions) msidIsStringNilOrBlank:]` only special-cased `nil` and `NSNull`. For any other non-`NSString` argument it fell through to `!string.length`, sending `-length` to the object:

```objc
if (!string || [string isKindOfClass:[NSNull class]] || !string.length)
```

When a mis-typed value reaches the check — e.g. a JSON/plist **boolean** decoded as `__NSCFBoolean` from an MDM SSO-extension configuration where a string was expected — this raises `-[__NSCFBoolean length]: unrecognized selector sent to instance` and crashes (SIGABRT).

This was observed as a crash in the iOS `AuthenticatorSSOExtension` on the main thread during `-[ASSSOExtensionAuthenticationViewController beginAuthorizationWithRequest:]`, where SSO-extension configuration values are parsed.

This change replaces the `NSNull`-only check with a positive `NSString` kind check, so any non-`NSString` (including `NSNull`, `NSNumber`/`__NSCFBoolean`, collections, etc.) is treated as blank and the method never messages a non-string:

```objc
if (!string || ![string isKindOfClass:[NSString class]] || !string.length)
```

Adds a regression test covering `NSNumber`/`__NSCFBoolean`, `NSArray` and `NSDictionary` inputs.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

## Additional information

Behavior for existing inputs is unchanged: `nil`, `NSNull`, empty and whitespace-only strings still return `YES`; non-blank strings still return `NO`. Only previously-crashing non-`NSString` inputs change from "throw" to `YES`. All `MSIDTestNSStringHelperMethods` tests pass locally (iOS, Xcode 26.5, iPhone 17 simulator), including the new regression test. Changelog updated under `TBD`.

### DIFF
--- a/IdentityCore/src/util/NSString+MSIDExtensions.m
+++ b/IdentityCore/src/util/NSString+MSIDExtensions.m
@@ -73,7 +73,10 @@ typedef unsigned char byte;
 
 + (BOOL)msidIsStringNilOrBlank:(NSString *)string
 {
-    if (!string || [string isKindOfClass:[NSNull class]] || !string.length)
+    // Treat any non-NSString (nil, NSNull, or a mis-typed value such as an NSNumber/__NSCFBoolean
+    // decoded from JSON or an MDM configuration plist) as blank, rather than sending -length to it
+    // and raising an unrecognized-selector exception.
+    if (!string || ![string isKindOfClass:[NSString class]] || !string.length)
     {
         return YES;
     }
--- a/IdentityCore/tests/MSIDStringExtensionsTests.m
+++ b/IdentityCore/tests/MSIDStringExtensionsTests.m
@@ -96,6 +96,18 @@
     XCTAssertFalse([NSString msidIsStringNilOrBlank:str], "Not an empty string %@", str);
 }
 
+- (void)testMsidIsStringNilOrBlank_whenNonStringObject_shouldReturnTrue
+{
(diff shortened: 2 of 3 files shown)


IdentityCore/src/util/NSString+MSIDExtensions.m
```diff
@@ -73,7 +73,10 @@ - (NSString *)msidBase64UrlDecode
 
 + (BOOL)msidIsStringNilOrBlank:(NSString *)string
 {
-    if (!string || [string isKindOfClass:[NSNull class]] || !string.length)
+    // Treat any non-NSString (nil, NSNull, or a mis-typed value such as an NSNumber/__NSCFBoolean
```
kaisong1990, line 76: Clean, behavior-preserving change — nil/NSNull/empty/whitespace still return YES, only the mistyped non-string case flips from a crash to YES. With ~270 call sites across the lib, hardening this one is well worth it. LGTM.

kaisong1990 (approved):

PR #1896, opened 2026-07-12
[patch] [engg]: Harden build pipelines + build.py formatter/coverage fixes

josephpab:

- [x] All tests pass locally

## Proposed changes

Brings common core's build pipelines in line with the recent MSAL and broker pipeline improvements. CI/build-script only — no runtime SDK code changes.

### CI reliability — silent-pass hardening
The `Run Build script` steps read `build.py`'s status from `build/status.txt` but never called `exit`, so a failed `build.py` could report a **green stage** (they relied solely on `failOnStderr: true`). All 7 build steps (`pr-validation` iOS/mac, `msal_submodule_check` iOS/mac, `visionos-validation` framework/broker, `broker_build_steps`) now:
- clear stale `status.txt`,
- capture `build.py`'s exit code as a fallback (for SIGKILL/timeout/early crash),
- emit `##vso[task.logissue type=error]` and **`exit` the real status**,
- set `failOnStderr: false` (failure now comes from the exit code, not stray tool stderr).

### build.py
- **xcbeautify** is now preferred for formatted output (graceful fallback to xcpretty, then raw), with JUnit via `xcbeautify --report junit`; adds a `--no-xcbeautify` alias. Safe on agents that only have xcpretty (it falls back).
- **Coverage gated on the real xcodebuild action** (`test`/`build-for-testing`/`test-without-building`) so passing `--coverage` with build-only schemes can't fail with *"only supported when testing"*.
- **Explicit `--target` alias** (CI invokes `--target`; previously worked only via argparse prefix abbreviation of `--targets`).

### Follow-ups (planned, not in this PR)
Once `IDDP/MSAL-ObjC-Pipelines` PR #25392 merges, adopt the shared step templates (`steps-publish-xcresults.yml`, `steps-crash-report.yml`) for the richer failed-test reporting + crash artifacts, rather than porting that logic into `build.py`.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High
- [x] Medium – CI infrastructure change affecting every build through these pipelines; the changes make failures *more* visible, not less, and build.py's formatter change falls back safely.
- [ ] Small

## Additional information

Mirrors the silent-pass fix + xcbeautify/coverage/`--target` changes already applied to the MSAL (`microsoft-authentication-library-for-objc` #2999) and broker pipelines.

### DIFF
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -21,7 +21,7 @@ resources:
     - repository: pipelinesShared
       type: git
       name: IDDP/MSAL-ObjC-Pipelines
-      ref: refs/heads/main
+      ref: refs/heads/josephpab/xcbeautify-and-shared-step-templates
 
 # Define parallel jobs that run build script for specified targets
 jobs:
--- a/azure_pipelines/pr-validation.yml
+++ b/azure_pipelines/pr-validation.yml
@@ -109,15 +109,21 @@ jobs:
         inputs:
           targetType: 'inline'
           script: |
+            rm -f ./build/status.txt
             { output=$(./build.py --no-clean --show-build-settings --target ios_library 2>&1 1>&3-) ;} 3>&1
-            final_status=$(<./build/status.txt)
+            build_exit=$?
+            final_status=$(cat ./build/status.txt 2>/dev/null || true)
+            if [ -z "${final_status}" ]; then
+              final_status="${build_exit}"
+            fi
             echo "FINAL STATUS  = ${final_status}"
             echo "POSSIBLE ERRORS: ${output}"
 
-            if [ $final_status != "0" ]; then
-              echo "Build & Testing Failed! \n ${output}" >&2
+            if [ "${final_status}" != "0" ]; then
+              echo "##vso[task.logissue type=error]build.py failed (status ${final_status}) - see 'POSSIBLE ERRORS' above"
+              exit "${final_status}"
             fi
-          failOnStderr: true
+          failOnStderr: false
       - task: Bash@3
         condition: always()
         displayName: Cleanup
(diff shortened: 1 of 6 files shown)


azure_pipelines/pr-validation.yml
```diff
@@ -21,7 +21,7 @@ resources:
     - repository: pipelinesShared
       type: git
       name: IDDP/MSAL-ObjC-Pipelines
-      ref: refs/heads/main
+      ref: refs/heads/josephpab/xcbeautify-and-shared-step-templates
```
mipetriu, line 24: These three flip the `IDDP/MSAL-ObjC-Pipelines` resource from `refs/heads/main` to `refs/heads/josephpab/xcbeautify-and-shared-step-templates`, and that repo isn't just declared — it's consumed live via `aces-macos-job.yml@pipelinesShared` (the job template every build extends). So merging this pins production CI on all three pipelines to a personal feature branch.

A couple of concerns:
~ That branch is mutable/deletable — once it's force-pushed or cleaned up, these pipelines stop resolving and every PR build breaks, with nothing in `dev` to point at. That's the kind of non-reproducible dependency we usually keep off `main`/`dev`.
~ It also seems to contradict the PR itself: the description says adopting the shared step templates is a follow-up "once IDDP/MSAL-ObjC-Pipelines PR #25392 merges," and the checklist has "no hidden dependencies" checked — but this is effectively that hidden dependency.

Is the repoint intentional for this PR (e.g. to get `xcbeautify` installed on the agents via the shared template), or is it a leftover from testing? If it's needed, could we hold this until #25392 lands on `main` and keep the ref at `refs/heads/main` here? If it's not needed, `build.py` already falls back to xcpretty/raw safely, so reverting these three back to `refs/heads/main` before merge would keep the PR independently mergeable.

antonioalwan:

PR #1900, opened 2026-07-14
Fix use-after-free race in MSIDFlightManager

DharshanBJ:

- [ ] All tests pass locally

## Proposed changes

Fixes a race condition in `MSIDFlightManager` that could cause a use-after-free crash (SIGSEGV) when flights are read concurrently with a provider swap.

Root cause: A data race / use-after-free on  _flightProvider . Readers accessed it through the unsynchronized nonatomic getter while  setFlightProvider:  used  dispatch_barrier_async , releasing the previous provider asynchronously. A concurrent reader could retain a pointer mid-release →  objc_retain  on freed memory → crash.

The fix ( IdentityCore/src/MSIDFlightManager.m ):

•  setFlightProvider:  →  dispatch_barrier_sync , so the assignment and release of the old provider complete before returning and are exclusive against reads.
•  boolForKey:  /  stringForKey:  (and the synchronized public  flightProvider  getter) read the ivar inside  dispatch_sync  into a strong local, then invoke the provider outside the block — keeping it alive during use and avoiding a re-entrancy deadlock.

Result: All reads/writes of  _flightProvider  are serialized on the queue with a strong reference held during use, eliminating the use-after-free.

Scope: `IdentityCore/src/MSIDFlightManager.m` (three methods).

> Note: this PR also contains an earlier commit adding a stray `// TEST` comment in `MSIDBrokerConstants.m`. That appears unintentional and can be dropped if not needed.

## Type of change

- [ ] Feature work
- [x] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk


IdentityCore/src/MSIDFlightManager.m
```diff
@@ -113,7 +113,11 @@ - (dispatch_queue_t)initializeDispatchQueue
 
 - (void)setFlightProvider:(id<MSIDFlightManagerInterface>)flightProvider
 {
-    dispatch_barrier_async(self.synchronizationQueue, ^{
+    // Use a synchronous barrier so the assignment (and the release of any previously
+    // installed provider) completes before the setter returns. An async barrier let the
+    // previous provider be released on the queue after the setter returned, racing with
+    // concurrent readers and allowing a stale/nil provider to be observed.
+    dispatch_barrier_sync(self.synchronizationQueue, ^{
```
kaisong1990, line 122: This can deadlock if a provider calls `setFlightProvider:` from `boolForKey:` or `stringForKey:`, since those callbacks run on the same queue and this uses `dispatch_barrier_sync`. Could we capture the provider while on the queue, then invoke it after leaving the queue?

DharshanBJ: updated,  boolForKey: / stringForKey:  now capture the provider under the queue and invoke it after leaving, so a re-entrant  setFlightProvider:  no longer deadlocks.

IdentityCore/src/MSIDFlightManager.m
```diff
@@ -137,36 +141,32 @@ - (void)setFlightProvider:(id<MSIDFlightManagerInterface>)flightProvider
...
-        id<MSIDFlightManagerInterface> flightProvider = self->_flightProvider;
-        if (flightProvider)
-        {
-            result = [flightProvider boolForKey:flightKey];
-        }
+        provider = self->_flightProvider;
     });
     
-    return result;
+    return provider ? [provider boolForKey:flightKey] : NO;
```
kaisong1990, line 154: The use-after-free path is already covered by `testConcurrentReads_whileProviderSwappedAndCleared_doNotCrash`, but the re-entrancy case this change fixes isn't covered anywhere. Can we add a regression test with a provider whose `boolForKey:` calls `setFlightProvider:` back on the same manager? It's deterministic now and would have hung before this change.

antonioalwan:

PR #1904, opened 2026-07-17
[Minot] [Feature]: Add trusted-host check before injecting custom headers in embedded webview

sidudiki:
## Summary
The embedded webview's `customHeaderProvider` path previously injected provider-returned headers for any HTTPS host, relying entirely on the provider for host trust. This adds an in-framework trusted-host check (`isAADPublicCloud:`) before invoking the header provider, consistent with the existing PkeyAuth handler behavior. Non-AAD hosts skip header injection and navigation proceeds normally; AAD hosts are unchanged.

## Changes
- Add a trusted-host gate in `MSIDOAuth2EmbeddedWebviewController` `decidePolicyForNavigationAction:` before calling `getCustomHeaders:forHost:`. Non-AAD host → skip the provider, inject no headers, and allow navigation (no cancel). AAD host → existing cancel + reload-with-headers behavior.
- Add `MSIDCustomHeaderTag` execution-flow telemetry (`Added` / `SkippedUntrustedHost`), mirroring the PkeyAuth handler tags.
- Add unit tests covering a trusted AAD host, another host, and a mid-navigation host change.

## Testing
- Mac and iOS test builds succeed.
- 38 tests pass (`MSIDOAuth2EmbeddedWebviewControllerTests` + `MSIDPKeyAuthHandlerTests`), 0 failures.


IdentityCore/src/webview/embeddedWebview/MSIDOAuth2EmbeddedWebviewController.m
```diff
@@ -545,6 +562,11 @@ - (void)decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction
                             [newUrlRequest setValue:extraHeaders[headerKey] forHTTPHeaderField:headerKey];
                         }
                     }
+
+                    if (self.context.correlationId)
+                    {
+                        MSIDExecutionFlowInsertTag(MSIDCustomHeaderTagToString(MSIDCustomHeaderAddedTag), nil, self.context.correlationId);
```
kaisong1990: Can we track whether at least one nonblank header was applied before canceling and reloading? A nonempty dictionary containing only blank values skips every `setValue:` call but still reloads the same request, so this can loop on each navigation and records `Added` even though nothing was attached. If no header was applied, the current navigation should continue.

IdentityCore/src/webview/embeddedWebview/MSIDOAuth2EmbeddedWebviewController.m
```diff
@@ -530,6 +532,21 @@ - (void)decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction
     
     if (self.customHeaderProvider)
     {
+        // Only forward custom headers to a known AAD host, consistent with the AAD host check used for other AAD requests.
+        NSString *requestHost = requestURL.host.lowercaseString;
+        if (![MSIDAADNetworkConfiguration.defaultConfiguration isAADPublicCloud:requestHost])
```
kaisong1990: Could we include hosts validated through instance metadata here? `isAADPublicCloud:` only checks the static list, while the cross-domain header provider also accepts `allCloudNetworkEnvironments`. As written, a valid discovered sovereign/private-cloud host is blocked before the provider can attach the matching PRT. A cached-host test would cover this path.

IdentityCore/src/webview/embeddedWebview/MSIDOAuth2EmbeddedWebviewController.m
```diff
@@ -530,6 +532,21 @@ - (void)decidePolicyForNavigationAction:(WKNavigationAction *)navigationAction
     
     if (self.customHeaderProvider)
     {
+        // Only forward custom headers to a known AAD host, consistent with the AAD host check used for other AAD requests.
+        NSString *requestHost = requestURL.host.lowercaseString;
```
jasoncoolmax, line 545: Solid, security-positive gate — forwarding the `MSID_REFRESH_TOKEN_CREDENTIAL` header only to exact-match trusted hosts (before `getCustomHeaders`) prevents credential-header leakage to untrusted hosts.

**Minor (test coverage):** the trust check depends on this `.lowercaseString` normalization, but no test exercises a mixed-case trusted host (e.g. `LOGIN.microsoftonline.com`). Consider adding one so a future regression dropping `.lowercaseString` (which would silently skip header injection for uppercase hosts) is caught.

sidudiki: Added testDecidePolicyForNavigationAction_whenMixedCaseAADHost_andProviderReturnsHeaders_shouldInjectHeadersAndReload to verifi mixed host

kaisong1990 (approved):

PR #1929, opened 2026-08-06
[minor] [feature]: Require bound refresh tokens for SPA silent acquisition

agubuzomaximus:

- [x] All tests pass locally

## PR Title Format

**Examples:**

## Proposed changes

This PR adds an opt-in `requiresBoundRefreshToken` constraint to the CommonCore silent token request. When enabled, access-token lookup remains unchanged, family refresh-token lookup is skipped, regular refresh tokens are rejected, and only a Bound App Refresh Token (BART) may be redeemed.

The Broker flow already distinguishes BART from regular refresh tokens when servicing this Browser Native Message scenario. The unmanaged flow will execute locally through CommonCore, so its silent acquisition must preserve the same security and routing guarantee rather than accidentally redeeming a regular RT or FRT when a BART is required. Exposing the token type through `isBoundRefreshToken` allows the shared silent engine to enforce that requirement without depending on Broker production code.

The new behavior defaults to disabled, so existing silent requests continue using their current RT and FRT behavior. A later unmanaged-flow PR will explicitly enable the constraint for local SPA acquisition.

This is the third independently mergeable phase extracted from #1875 and builds on the shared routing and request-mapping changes introduced in #1927 and #1928.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High - Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [x] Medium - Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
- [ ] Small - No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)

## Additional information

- iOS: 39/39 focused refresh-token and silent-request tests passed.
- macOS: 39/39 focused refresh-token and silent-request tests passed.
- Dedicated coverage verifies default-off behavior, regular RT rejection, BART acceptance, and FRT lookup suppression.

### DIFF
--- a/IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
+++ b/IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
@@ -169,6 +169,11 @@
 
 - (nullable MSIDRefreshToken *)familyRefreshTokenWithError:(NSError * _Nullable __autoreleasing * _Nullable)error
 {
+    if (self.requiresBoundRefreshToken)
+    {
+        return nil;
+    }
+
     self.appMetadata = [self appMetadataWithError:error];
     self.defaultAccessor.shouldSkipBoundAppRefreshTokenLookup = self.shouldSkipBoundAppRefreshTokenUsage;
 
@@ -190,11 +195,18 @@
 - (nullable MSIDBaseToken<MSIDRefreshableToken> *)appRefreshTokenWithError:(NSError * _Nullable __autoreleasing * _Nullable)error
 {
     self.defaultAccessor.shouldSkipBoundAppRefreshTokenLookup = self.shouldSkipBoundAppRefreshTokenUsage;
-    return [self.defaultAccessor getRefreshTokenWithAccount:self.requestParameters.accountIdentifier
-                                                   familyId:nil
-                                              configuration:self.requestParameters.msidConfiguration
-                                                    context:self.requestParameters
-                                                      error:error];
+    MSIDRefreshToken *refreshToken = [self.defaultAccessor getRefreshTokenWithAccount:self.requestParameters.accountIdentifier
+                                                                             familyId:nil
+                                                                        configuration:self.requestParameters.msidConfiguration
+                                                                              context:self.requestParameters
+                                                                                error:error];
+    if (self.requiresBoundRefreshToken && refreshToken && !refreshToken.isBoundRefreshToken)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters, @"Ignoring regular refresh token because a bound refresh token is required.");
+        return nil;
+    }
+
+    return refreshToken;
 }
 
 - (BOOL)updateFamilyIdCacheWithServerError:(NSError *)serverError
--- a/IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
+++ b/IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
@@ -171,6 +171,7 @@
 {
     if (self.requiresBoundRefreshToken)
     {
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters, @"Skipping family refresh token lookup because an app-specific bound refresh token is required.");
         return nil;
     }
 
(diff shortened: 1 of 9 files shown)


IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
```diff
@@ -169,6 +169,11 @@ -(MSIDIdToken *)getIDTokenForTokenType:(MSIDCredentialType)idTokenType
 
 - (nullable MSIDRefreshToken *)familyRefreshTokenWithError:(NSError * _Nullable __autoreleasing * _Nullable)error
 {
+    if (self.requiresBoundRefreshToken)
```
kaisong1990, line 172: Why is the family path an unconditional bail-out, while `appRefreshTokenWithError:` below fetches first and then filters on `isBoundRefreshToken`? `getRefreshTokenWithAccount:familyId:` forwards `familyId` straight into the BART lookup, so a bound FRT is representable in the cache - this drops it and forces interaction_required even when a valid bound token exists.

Suggest mirroring the app RT path: fetch, then return nil only when `!refreshToken.isBoundRefreshToken`, and log the reason like that branch does. Right now the family path returns nil silently, which is hard to diagnose in the field.

antrix1989: can Bound RT be FRT at the same time?

agubuzomaximus: Hi Kai. this flow only allows for bound RT's. 
I can log here when we return nil. I can even create an error interaction required.

If nil is returend we fall back into interactive path (flip to authenticator, mint a PRT for the client id, creating a BART and AT) then return back OneAuth. 

There are two production code paths that create  MSIDBoundRefreshToken , but they are construction paths rather than separate user flows:

 1.  MSIDOauth2Factory.m:174-176  — while creating an RT from a server response, it sees a BART device ID and wraps the normal RT as  MSIDBoundRefreshToken .
 2.  MSIDTokenResponseValidator.m:283-286  — after validating a response, it sees  refresh_token_type=bound_app_rt  plus a device ID and wraps the result RT as  MSIDBoundRefreshToken .

Only  MSIDBoundRefreshToken  returns:

 isBoundRefreshToken == YES

A normal  MSIDRefreshToken  returns  NO .  MSIDFamilyRefreshToken  inherits from  MSIDRefreshToken  and does not override the property, so an ordinary FRT returns  NO .

agubuzomaximus: Hi @antrix1989 there is no production flow that marks FRT with bound app refresh token.

IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
```diff
@@ -190,11 +196,18 @@ - (nullable MSIDRefreshToken *)familyRefreshTokenWithError:(NSError * _Nullable
...
-                                                    context:self.requestParameters
-                                                      error:error];
+    MSIDRefreshToken *refreshToken = [self.defaultAccessor getRefreshTokenWithAccount:self.requestParameters.accountIdentifier
+                                                                             familyId:nil
+                                                                        configuration:self.requestParameters.msidConfiguration
+                                                                              context:self.requestParameters
+                                                                                error:error];
+    if (self.requiresBoundRefreshToken && refreshToken && !refreshToken.isBoundRefreshToken)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, self.requestParameters, @"Ignoring regular refresh token because a bound refresh token is required.");
```
rokhrest, line 206: nit: a little weird log line because it's possible the RT returned is nil. But if it's just a copypaste from old log, then it's fine. :)

rokhrest (approved):

IdentityCore/src/requests/sdk/msal/MSIDDefaultSilentTokenRequest.m
agubuzomaximus: Hello sir, returning nil here is okay as well because we then flip into an interactive flow which will produce the artifacts we need for a silent flow on the next request.

PR #1931, opened 2026-08-07
[patch][engg]: Add iOS 27 NSURL percent double-encode regression tests

mipetriu:
### Summary

Adds **iOS/macOS 27 regression tests only** for the SDK 27 `NSURL` percent-encoding behavior change, where `NSURL` no longer double-encodes `%` in already-valid percent-escape sequences.

IdentityCore's URL construction relies on `percentEncodedQuery` / `percentEncodedHost` and the custom `msidURLEncode` helper explicitly to avoid double-encoding. These paths touch every Microsoft app's login-URL construction, so this locks in the expected single-encode behavior with explicit tests. **No production/src code is changed** — the existing implementation already aligns with the new 27 behavior; these tests guard against regressions.

### Changes

- `IdentityCore/tests/MSIDURLExtensionsTests.m` — 3 new tests on `msidURLWithQueryParameters:`:
  - pre-encoded value in an existing query is **not** double-encoded
  - an appended value containing a literal `%` is encoded exactly once (`%` → `%25`)
  - an appended value that *looks like* an escape (`%20`) has its `%` encoded literally
- `IdentityCore/tests/MSIDStringExtensionsTests.m` — 2 new tests on `msidURLEncode` / dictionary encoding:
  - literal `%` → `%25`
  - reserved chars in dictionary values encoded once (uses `%20`, not `+`)

### Validation

Built and ran on a real **iOS 27.0 simulator** (iPhone 17, runtime 24A5390f):

```
xcodebuild test -workspace IdentityCore.xcworkspace -scheme "IdentityCore iOS" \
  -destination 'platform=iOS Simulator,name=iPhone 17,OS=27.0' \
  -only-testing:"IdentityCoreTests iOS/NSURLExtensionsTests" \
  -only-testing:"IdentityCoreTests iOS/MSIDTestNSStringHelperMethods"
** TEST SUCCEEDED **  — 90 executed, 0 failures
```

(Branch includes the merged MSIDTestsHostApp UIScene migration from `dev`, so the IdentityCore iOS test bundle launches cleanly under the iOS 27 runtime.)

### Risk

Tests-only; no runtime/behavioral change. `[patch][engg]` — changelog-exempt.

### DIFF
--- a/IdentityCore/tests/MSIDStringExtensionsTests.m
+++ b/IdentityCore/tests/MSIDStringExtensionsTests.m
@@ -279,7 +279,7 @@
     XCTAssertEqualObjects(result, @"key=");
 }
 
-- (void)testMsidURLEncode_when27_hasLiteralPercent_shouldEncodeToPercent25
+- (void)testMsidURLEncode_whenLiteralPercent_shouldEncodeToPercent25
 {
     // iOS/macOS 27 (radar 161588649): msidURLEncode escapes a literal % to %25 so the non-re-encoding setter stays correct.
     XCTAssertEqualObjects([@"50%off" msidURLEncode], @"50%25off");
@@ -287,7 +287,7 @@
     XCTAssertEqualObjects([@"a%20b" msidURLEncode], @"a%2520b");
 }
 
-- (void)testMsidURLEncodedStringFromDictionary_when27_valueHasReservedChars_shouldEncodeOnce
+- (void)testMsidURLEncodedStringFromDictionary_whenValueHasReservedChars_shouldEncodeOnce
 {
     // iOS/macOS 27 (radar 161588649): reserved chars in a redirect_uri-style value are encoded exactly once.
     NSString *encoded = [NSString msidURLEncodedStringFromDictionary:@{@"redirect_uri": @"msauth://com.example/cb"}];
--- a/IdentityCore/tests/MSIDStringExtensionsTests.m
+++ b/IdentityCore/tests/MSIDStringExtensionsTests.m
@@ -290,9 +290,9 @@
 - (void)testMsidURLEncodedStringFromDictionary_whenValueHasReservedChars_shouldEncodeOnce
 {
     // iOS/macOS 27 (radar 161588649): reserved chars in a redirect_uri-style value are encoded exactly once.
-    NSString *encoded = [NSString msidURLEncodedStringFromDictionary:@{@"redirect_uri": @"msauth://com.example/cb"}];
-    XCTAssertEqualObjects(encoded, @"redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb");
-    XCTAssertFalse([encoded containsString:@"%253A"]);
+    // The space pins the non-form encoder: msidWWWFormURLEncodedStringFromDictionary would emit '+' instead of %20.
+    NSString *encoded = [NSString msidURLEncodedStringFromDictionary:@{@"redirect_uri": @"msauth://com.example/cb path"}];
+    XCTAssertEqualObjects(encoded, @"redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb%20path");
 }
 
 - (void)testmsidJson_whenNotJson_shouldReturnNil
--- a/IdentityCore/tests/MSIDStringExtensionsTests.m
+++ b/IdentityCore/tests/MSIDStringExtensionsTests.m
@@ -279,6 +279,22 @@
     XCTAssertEqualObjects(result, @"key=");
 }
 
+- (void)testMsidURLEncode_when27_hasLiteralPercent_shouldEncodeToPercent25
+{
+    // iOS/macOS 27 (radar 161588649): msidURLEncode escapes a literal % to %25 so the non-re-encoding setter stays correct.
+    XCTAssertEqualObjects([@"50%off" msidURLEncode], @"50%25off");
+    // A pre-existing escape in raw input is treated as literal text: % -> %25.
+    XCTAssertEqualObjects([@"a%20b" msidURLEncode], @"a%2520b");
+}
+
+- (void)testMsidURLEncodedStringFromDictionary_when27_valueHasReservedChars_shouldEncodeOnce
+{
+    // iOS/macOS 27 (radar 161588649): reserved chars in a redirect_uri-style value are encoded exactly once.
+    NSString *encoded = [NSString msidURLEncodedStringFromDictionary:@{@"redirect_uri": @"msauth://com.example/cb"}];
+    XCTAssertEqualObjects(encoded, @"redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb");
+    XCTAssertFalse([encoded containsString:@"%253A"]);
+}
+
 - (void)testmsidJson_whenNotJson_shouldReturnNil
 {
     NSString *jsonString = @"{\"-not\":a json&*";
--- a/IdentityCore/tests/MSIDURLExtensionsTests.m
+++ b/IdentityCore/tests/MSIDURLExtensionsTests.m
@@ -378,7 +378,7 @@
     XCTAssertEqualObjects(resultURL, expectedResultURL);
 }
 
-- (void)testMsidURLWithQueryParameters_when27_existingQueryHasPreEncodedValue_shouldNotDoubleEncode
+- (void)testMsidURLWithQueryParameters_whenExistingQueryHasPreEncodedValue_shouldNotDoubleEncode
 {
     // iOS/macOS 27 (radar 161588649): NSURL must not re-encode an already-percent-encoded existing query.
     NSURL *inputURL = [NSURL URLWithString:@"https://somehost.com:652?redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb"];
@@ -389,7 +389,7 @@
     XCTAssertFalse([resultURL.absoluteString containsString:@"%252F"]);
 }
 
-- (void)testMsidURLWithQueryParameters_when27_appendedValueHasLiteralPercent_shouldEncodeOnce
+- (void)testMsidURLWithQueryParameters_whenAppendedValueHasLiteralPercent_shouldEncodeOnce
 {
     // iOS/macOS 27 (radar 161588649): a literal % in an appended value must be encoded exactly once (%25), never doubled.
     NSURL *inputURL = [NSURL URLWithString:@"https://somehost.com:652?existing1=value2"];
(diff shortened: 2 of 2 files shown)


IdentityCore/tests/MSIDStringExtensionsTests.m
```diff
@@ -279,6 +279,22 @@ - (void)testmsidWWWFormURLEncodedStringFromDictionary_whenKeyWithEmptyValue_shou
...
+    // iOS/macOS 27 (radar 161588649): msidURLEncode escapes a literal % to %25 so the non-re-encoding setter stays correct.
+    XCTAssertEqualObjects([@"50%off" msidURLEncode], @"50%25off");
+    // A pre-existing escape in raw input is treated as literal text: % -> %25.
+    XCTAssertEqualObjects([@"a%20b" msidURLEncode], @"a%2520b");
+}
+
+- (void)testMsidURLEncodedStringFromDictionary_when27_valueHasReservedChars_shouldEncodeOnce
+{
+    // iOS/macOS 27 (radar 161588649): reserved chars in a redirect_uri-style value are encoded exactly once.
+    NSString *encoded = [NSString msidURLEncodedStringFromDictionary:@{@"redirect_uri": @"msauth://com.example/cb"}];
```
kaisong1990: This value has no space, so it passes identically for msidWWWFormURLEncodedStringFromDictionary and does not actually pin the %20-vs-+ behavior the test name implies. Suggest `@"msauth://com.example/cb path"` and asserting `%20` so the non-form encoder is distinguished.

IdentityCore/tests/MSIDURLExtensionsTests.m
```diff
@@ -378,6 +378,34 @@ - (void)testMsidURLWithQueryParameters_whenNonEmptyQuery_NonEmptyParametersWithS
...
 
+- (void)testMsidURLWithQueryParameters_when27_existingQueryHasPreEncodedValue_shouldNotDoubleEncode
+{
+    // iOS/macOS 27 (radar 161588649): NSURL must not re-encode an already-percent-encoded existing query.
+    NSURL *inputURL = [NSURL URLWithString:@"https://somehost.com:652?redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb"];
+    NSURL *resultURL = [inputURL msidURLWithQueryParameters:@{@"key2": @"value2"}];
+    XCTAssertEqualObjects(resultURL.absoluteString, @"https://somehost.com:652?redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb&key2=value2");
+    // Guard: the pre-encoded %3A/%2F must NOT become %253A/%252F.
+    XCTAssertFalse([resultURL.absoluteString containsString:@"%253A"]);
+    XCTAssertFalse([resultURL.absoluteString containsString:@"%252F"]);
```
kaisong1990: These two containsString guards are already covered by the exact absoluteString equality above - if %253A/%252F appeared the equality assert would fail first. Same applies to the %2525 guard in the next test. Can drop them to keep the tests to one assertion each.

IdentityCore/tests/MSIDURLExtensionsTests.m
```diff
@@ -378,6 +378,34 @@ - (void)testMsidURLWithQueryParameters_whenNonEmptyQuery_NonEmptyParametersWithS
...
+}
```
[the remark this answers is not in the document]
mipetriu: Thanks — good catch on the naming. I've renamed all five to drop the `when27_` segment (`30838ff`); the radar reference now lives only in the inline comments.

On the gating suggestion, I'd push back. These tests assert invariants that should hold **identically on every OS version**. The premise of this change is that IdentityCore's `msidURLEncode` workarounds already aligned with iOS 27's non-double-encoding behavior, so the expected outputs are the same before and after. Running unconditionally on old and new runtimes is exactly what proves that compatibility — gating to 27+ would discard the older-OS regression signal and the cross-version equivalence proof, i.e. strictly reduce coverage.

Also worth noting only one of the five actually touches changed API: `msidURLWithQueryParameters` (via `percentEncodedQuery`). `msidURLEncode` / `msidURLEncodedStringFromDictionary` are our own category methods over `stringByAddingPercentEncodingWithAllowedCharacters`, which is version-stable.

Verified empirically — identical assertions, 90 tests / 0 failures on all three runtimes, with all 5 of these passing on each:

| Runtime | Device | Result |
|---|---|---|
| iOS 27.0 (24A5390f) | iPhone 17 | 90 tests, 0 failures |
| iOS 26.5 (23F77) | iPhone 17 | 90 tests, 0 failures |
| iOS 17.5 (21F79) | iPhone 15 Pro | 90 tests, 0 failures |

Resolving as addressed-by-rename, not-gating.

IdentityCore/tests/MSIDStringExtensionsTests.m
```diff
@@ -279,6 +279,22 @@ - (void)testmsidWWWFormURLEncodedStringFromDictionary_whenKeyWithEmptyValue_shou
     XCTAssertEqualObjects(result, @"key=");
```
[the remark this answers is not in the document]
mipetriu: Working as intended — this PR is test-only (adds regression coverage for the iOS/macOS 27 NSURL percent-encoding change) and carries the `[patch][engg]` prefix, so it is changelog-exempt. No partner- or user-visible behavior change.

IdentityCore/tests/MSIDStringExtensionsTests.m
mipetriu: Good catch — you're right that the value had no space, so `msidURLEncodedStringFromDictionary` and `msidWWWFormURLEncodedStringFromDictionary` produced byte-identical output and the test didn't actually pin the non-form encoder. (They differ by exactly one transform: `msidWWWFormURLEncode` is `msidURLEncode` with `%20` -> `+`.)

Fixed in `00fd617` — value is now `msauth://com.example/cb path`, asserting `redirect_uri=msauth%3A%2F%2Fcom.example%2Fcb%20path`. The form encoder would emit `+` there, so the two are now distinguished. Added a comment saying so, and dropped the redundant `%253A` guard per your other comment.

Verified passing on iOS 27.0, 26.5, and 17.5.

IdentityCore/tests/MSIDURLExtensionsTests.m
mipetriu: Agreed — the exact `absoluteString` equality subsumes all three guards; if double-encoding occurred the equality assert would fail first. Removed in `00fd617`: both `%253A`/`%252F` guards (and their `// Guard:` comment) from the pre-encoded test, and the `%2525` guard from the literal-percent test. Also dropped the equivalent `%253A` guard in the `MSIDStringExtensionsTests` case for consistency.

`grep -n containsString IdentityCore/tests/MSIDURLExtensionsTests.m` now returns nothing.

Verified passing on iOS 27.0, 26.5, and 17.5 (90 tests, 0 failures on each).

IdentityCore/tests/MSIDStringExtensionsTests.m
```diff
@@ -279,6 +279,22 @@ - (void)testmsidWWWFormURLEncodedStringFromDictionary_whenKeyWithEmptyValue_shou
...
 }
 
+- (void)testMsidURLEncode_whenLiteralPercent_shouldEncodeToPercent25
```
jasoncoolmax, line 282: Good coverage. One thing while we are here: `MSIDRequestParameters.m:170` does `tokenEndpoint.query = [endpointQPs msidURLEncode];`, i.e. it feeds an already percent-encoded string into the `.query` setter instead of `percentEncodedQuery`. Every other call site (`MSIDWebviewFactory`, `MSIDCertAuthManager`, `MSIDPKeyAuthHandler`, `MSIDSwitchBrowserOperation`, `MSIDUrlRequestSerializer`) uses `percentEncodedQuery`. I tried it locally and the `.query` setter already double encodes today (`%3A` -> `%253A`), so `extraURLQueryParameters` with reserved chars looks broken. Shall we cover that one too?

mipetriu: Good catch — confirmed, and it's worse than "inconsistent style". I dug into it before replying:

**It's a real double-encode.** Verified empirically:

```
u.query               = "a=x%3Ay"  ->  a=x%253Ay   // re-encodes the '%'
u.percentEncodedQuery = "a=x%3Ay"  ->  a=x%3Ay     // round-trips
```

Since line 157 reads `percentEncodedQuery` and *decodes* it, and 166 merges in raw `extraURLQueryParameters`, everything reaching 170 is already encoded once by `msidURLEncode` — then the `.query` setter encodes it a second time.

**It's the only site doing this.** Six places feed `msidURLEncode` output into a query setter; five already use `percentEncodedQuery` (`MSIDUrlRequestSerializer.m:49`, `MSIDPKeyAuthHandler.m:81`, `MSIDSwitchBrowserOperation.m:94`, `MSIDCertAuthManager.m:109`, `MSIDWebviewFactory.m:366`). Every *other* `.query =` in the repo assigns `nil`, which is just query-stripping and harmless. So `MSIDRequestParameters.m:170` is the lone outlier.

**It's a missed migration, not a deliberate choice.** `git blame` puts line 170 at 2019-02-01; the five correct call sites landed later (2019-11-22 and 2020-03-25). The codebase moved to `percentEncodedQuery` and this one didn't come along.

**Why it's gone unnoticed for 7 years:** it only manifests when a value contains a reserved character. Typical AAD token endpoints have an empty query, and the extra params we see in practice (`slice=testslice`, `dc=ESTS-PUB-...`) contain nothing `msidURLEncode` would escape, so the setter is a no-op passthrough. Every existing test uses reserved-char-free values too (`value1`, `val1`) — which also means no test currently pins the buggy output, so a fix won't break anything.

---

**On whether it belongs in this PR — I'd like to keep it out, and here's the specific reason.**

I initially assumed this might be in scope because of the iOS 27 `NSURL` percent-encoding change (161588649). It isn't. I built the test above against the iOS 27.0 SDK and ran it on a real iOS 27.0 simulator (build `24A5390f`, runtime version printed to be sure it wasn't silently the host):

```
RUNTIME: 27.0.0
.query                -> a=x%253Ay
.percentEncodedQuery  -> a=x%3Ay
```

Identical to macOS 26. The 27 change doesn't touch the `.query` setter, so this is a plain pre-existing bug rather than a 27 regression.

That matters for scoping: this PR is test-only (+40/−0) and deliberately narrow to 27 behavior. A test covering line 170 here would either fail — blocking an otherwise green PR — or have to assert the double-encoded string, which would pin the bug in place.

So I'd rather file it separately: a one-line fix (`.query` -> `.percentEncodedQuery`) plus a regression test with a reserved character in the value. Worth prioritising properly, since `tokenEndpoint` feeds every token request path (auth-code, refresh-token, bound-RT, nonce) and `extraURLQueryParameters` is reachable from public MSAL API (`MSALTokenParameters.extraQueryParameters`).

Happy to open that one right after this merges — or now in parallel if you'd prefer.

antonioalwan:
No additional high-confidence correctness, security, concurrency, API-contract, or test-coverage issues were found.

PR #1932, opened 2026-08-10
[minor] [feature]: Add local SPA silent token acquisition backend

agubuzomaximus:
## PR Checklist (must be completed before review)

- [x] All tests pass locally
- [x] PR size is <= 500 LOC per PR Size Check policy
- [x] PR is independently mergeable (no hidden dependencies)
- [ ] Appropriate reviewers are assigned
- [ ] PR reviewed by code owner (required if Copilot-generated)
- [ ] SME or Senior IC assigned where required

## PR Title Format

**Required Format:** `[Keyword1] [Keyword2]: Description`

- **Keyword1:** `major`, `minor`, or `patch` (case-insensitive)
- **Keyword2:** `feature`, `bugfix`, `engg`, or `tests` (case-insensitive)

**Examples:**
- `[MAJOR] [Feature]: new API`
- `[minor] [bugfix]: fix crash`
- `[PATCH][tests]:add coverage`

## Proposed changes

This PR adds the local silent-acquisition backend for the unmanaged Browser Native Message SPA flow. It maps the incoming GetToken request to broker-compatible interactive parameters with BART requested, builds the silent engine over the shared ADAL keychain caches, requires an app-specific bound refresh token, and returns a normalized `MSIDTokenResult` outcome.

In the existing Broker flow, the SSO Extension owns this silent redemption. Unmanaged iOS Edge cannot invoke that extension, so OneAuth must perform the same cached BART redemption in-process through CommonCore. Keeping the engine and cache wiring in `MSIDLocalSPATokenAcquirer` allows the next provider-orchestration phase to share routing and response shaping without depending on Broker production code.

This phase intentionally contains only silent acquisition. Shared routing, interactive fallback orchestration, and serialized response construction remain outside this backend and will be wired by the final OneAuth-facing provider PR.

This is the fifth independently mergeable phase extracted from #1875. It builds on the request mapping and app-specific BART enforcement introduced in #1928 and #1929; its normalized result will be consumed by the response shaping introduced in #1930.

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High - Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [x] Medium - Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
[...]

PR #1933, opened 2026-08-10
[minor] [feature]: Orchestrate unmanaged BART SPA token acquisition

agubuzomaximus:

- [x] All tests pass locally

## PR Title Format

**Examples:**

## Proposed changes

This PR replaces the `MSIDBoundTokenProvider` proof-of-concept payload with the OneAuth-facing orchestration for unmanaged Browser Native Message BART SPA acquisition. The provider validates and maps the request, applies the shared GetToken routing policy, delegates silent or interactive acquisition through `MSIDSPATokenAcquiring`, and serializes the normalized token result through `MSIDBrowserNativeMessageGetTokenResponse`.

On unmanaged iOS, OneAuth cannot invoke the SSO Extension. The provider therefore uses `MSIDLocalSPATokenAcquirer` for in-process access-token cache lookup and app-specific BART redemption. When silent acquisition requires interaction, the provider reruns routing with `forceInteractive` and invokes the interactive backend only when the request permits UI. The default local interactive backend currently returns interaction-required; the broker-flip implementation remains a separate integration step.

This PR also addresses Antonio's review comment from #1932: an empty account username is now treated as missing and falls back to the request login hint, preventing the Browser Native Message response from omitting `account.userName` and `properties.UPN`.

This is the sixth independently mergeable phase extracted from #1875. It composes the routing policy (#1927), request mapping (#1928), app-specific BART enforcement (#1929), response shaping (#1930), and local silent backend (#1932).

## Type of change

- [x] Feature work
- [ ] Bug fix
- [ ] Documentation
- [ ] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High - Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [x] Medium - Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
- [ ] Small - No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)

## Additional information

- iOS: 15/15 focused provider and local-acquirer tests passed.
- macOS: 15/15 focused provider and local-acquirer tests passed.
- Coverage verifies silent success, direct interactive routing, one-time silent-to-interactive fallback, UI-blocked behavior, hard-error propagation, request validation/mapping failures, Browser Native Message response shaping, and blank-username fallback.

### DIFF
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -203,6 +203,10 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
 {
     if (route != MSIDBrowserNativeMessageGetTokenRouteInteractive)
     {
+        NSString *routeReason = route == MSIDBrowserNativeMessageGetTokenRouteUIBlocked ? @"canShowUI is NO" : @"prompt=none forbids UI";
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
+                          @"%@ Interactive fallback rejected. Route: %ld. Reason: %@. promptType: %ld, canShowUI: %@.",
+                          MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, (long)route, routeReason, (long)parameters.promptType, request.canShowUI ? @"YES" : @"NO");
         NSError *error = interactionRequiredError
             ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
                                @"Bound token acquisition requires user interaction.",
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -198,10 +198,11 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
                           @"%@ Interactive fallback rejected. Route: %ld. Reason: %@. promptType: %ld, canShowUI: %@.",
                           MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, (long)route, routeReason, (long)parameters.promptType, request.canShowUI ? @"YES" : @"NO");
-        NSError *error = interactionRequiredError
-            ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
-                               @"Bound token acquisition requires user interaction.",
-                               nil, nil, nil, context.correlationId, nil, YES);
+        NSString *status = route == MSIDBrowserNativeMessageGetTokenRouteUIBlocked ? @"UI_NOT_ALLOWED" : @"USER_INTERACTION_REQUIRED";
+        NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
+                                         route == MSIDBrowserNativeMessageGetTokenRouteUIBlocked ? @"User interaction is required, but showing UI is not allowed." : @"User interaction is required, but prompt=none forbids UI.",
+                                         nil, nil, interactionRequiredError, context.correlationId,
+                                         @{@"MSALBrowserNativeMessageErrorStatus": status}, YES);
         completionBlock(nil, error);
         return;
     }
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -37,9 +37,7 @@
 NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
 
 @interface MSIDBoundTokenProvider ()
-
 @property (nonatomic) id<MSIDSPATokenAcquiring> acquirer;
-
 @end
 
 @implementation MSIDBoundTokenProvider
@@ -70,12 +68,10 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"%@ completionBlock is nil.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
         return;
     }
-
     if (![self validateRequest:request context:context completionBlock:completionBlock])
     {
         return;
     }
-
     NSError *parametersError = nil;
     MSIDInteractiveTokenRequestParameters *parameters =
     [self.acquirer requestParametersForRequest:request context:context error:&parametersError];
@@ -88,7 +84,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         completionBlock(nil, error);
         return;
     }
-
     MSIDBrowserNativeMessageGetTokenRoute route =
     [[self routingPolicy] routeWithForceInteractive:NO
                                          promptType:parameters.promptType
@@ -103,7 +98,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
                             completionBlock:completionBlock];
         return;
     }
-
     [self completeInteractiveRoute:route
                         parameters:parameters
                            request:request
@@ -135,7 +129,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         completionBlock(nil, error);
         return NO;
     }
-
     if ([NSString msidIsStringNilOrBlank:request.clientId] ||
         [NSString msidIsStringNilOrBlank:request.redirectUri])
     {
@@ -145,7 +138,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         completionBlock(nil, error);
         return NO;
     }
-
     return YES;
 }
 
@@ -167,7 +159,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
                       completionBlock:completionBlock];
             return;
         }
-
         BOOL interactionRequired = [error.domain isEqualToString:MSIDErrorDomain]
             && error.code == MSIDErrorInteractionRequired;
         if (interactionRequired)
@@ -186,7 +177,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
                   interactionRequiredError:error];
             return;
         }
-
         [self completeWithOutcome:nil
                             error:error
                           request:request
@@ -215,7 +205,6 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         completionBlock(nil, error);
         return;
     }
-
     [self.acquirer acquireInteractiveWithParameters:parameters
                                             request:request
                                             context:context
@@ -239,43 +228,45 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
         NSError *completionError = error
             ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
                                @"Bound token acquisition completed without a result or error.",
-                               nil, nil, nil, context.correlationId, nil, NO);
+                               nil, nil, nil, context.correlationId, nil, YES);
         completionBlock(nil, completionError);
         return;
     }
-
     NSString *fallbackUpn = outcome.fallbackRequestAccountUpn;
     if ([NSString msidIsStringNilOrBlank:fallbackUpn])
     {
         fallbackUpn = request.loginHint;
     }
-
     MSIDBrowserNativeMessageGetTokenResponse *response =
     [[MSIDBrowserNativeMessageGetTokenResponse alloc] initWithTokenResult:outcome.tokenResult
                                                                     state:request.state
                                                 fallbackRequestAccountUpn:fallbackUpn];
-    NSDictionary *responseDictionary = [response jsonDictionary];
-    if (!responseDictionary)
-    {
-        NSError *responseError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
-                                                 @"Failed to create bound token response.",
-                                                 nil, nil, nil, context.correlationId, nil, NO);
-        completionBlock(nil, responseError);
-        return;
-    }
-
+    NSDictionary *responseDictionary = response ? [response jsonDictionary] : nil;
     NSError *serializationError = nil;
-    NSData *data = [NSJSONSerialization dataWithJSONObject:responseDictionary options:0 error:&serializationError];
-    if (!data)
+    NSData *data = responseDictionary ? [NSJSONSerialization dataWithJSONObject:responseDictionary options:0 error:&serializationError] : nil;
+    NSString *responseString = data ? [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding] : nil;
+    if (!responseString)
     {
-        NSError *responseError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
-                                                 @"Failed to serialize bound token response.",
-                                                 nil, nil, serializationError, context.correlationId, nil, NO);
+        NSString *description = @"Failed to encode bound token response.";
+        if (!response)
+        {
+            description = @"Failed to initialize bound token response.";
+        }
+        else if (!responseDictionary)
+        {
+            description = @"Failed to create bound token response.";
+        }
+        else if (!data)
+        {
+            description = @"Failed to serialize bound token response.";
+        }
+        NSError *responseError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal, description,
+                                                 nil, nil, serializationError,
+                                                 context.correlationId, nil, YES);
         completionBlock(nil, responseError);
         return;
     }
-
-    completionBlock([[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding], nil);
+    completionBlock(responseString, nil);
 }
 
 @end
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -51,6 +51,7 @@ NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider
 
 - (instancetype)initWithAcquirer:(id<MSIDSPATokenAcquiring>)acquirer
 {
+    NSParameterAssert(acquirer);
     self = [super init];
     if (self)
     {
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -23,42 +23,104 @@
 // THE SOFTWARE.
 
 #import "MSIDBoundTokenProvider.h"
+#import "MSIDLocalSPATokenAcquirer.h"
 #import "MSIDBrowserNativeMessageGetTokenRequest.h"
+#import "MSIDBrowserNativeMessageGetTokenRoutingPolicy.h"
+#import "MSIDBrowserNativeMessageGetTokenResponse.h"
+#import "MSIDInteractiveTokenRequestParameters.h"
+#import "MSIDSPATokenAcquisitionResult.h"
+#import "MSIDDIContainer.h"
 #import "MSIDError.h"
 #import "MSIDLogger+Internal.h"
 #import "NSString+MSIDExtensions.h"
 
 NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
 
+@interface MSIDBoundTokenProvider ()
+
+@property (nonatomic) id<MSIDSPATokenAcquiring> acquirer;
+
+@end
+
 @implementation MSIDBoundTokenProvider
+
+- (instancetype)init
+{
+    return [self initWithAcquirer:[MSIDLocalSPATokenAcquirer new]];
+}
+
+- (instancetype)initWithAcquirer:(id<MSIDSPATokenAcquiring>)acquirer
+{
+    self = [super init];
+    if (self)
+    {
+        _acquirer = acquirer;
+    }
+    return self;
+}
 
 - (void)acquireBoundTokenWithRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
                              context:(nullable id<MSIDRequestContext>)context
                      completionBlock:(MSIDBoundTokenProviderCompletionBlock)completionBlock
 {
     NSParameterAssert(completionBlock);
-    if (!completionBlock) return;
+    if (!completionBlock)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"%@ completionBlock is nil.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
+        return;
+    }
 
     if (![self validateRequest:request context:context completionBlock:completionBlock])
     {
         return;
     }
 
-    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
-                      @"%@ Servicing GetToken request in-process (no SSO extension). clientId: %@",
-                      MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, request.clientId);
+    NSError *parametersError = nil;
+    MSIDInteractiveTokenRequestParameters *parameters =
+    [self.acquirer requestParametersForRequest:request context:context error:&parametersError];
+    if (!parameters)
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"%@ Failed to build request parameters: %@", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, MSID_PII_LOG_MASKABLE(parametersError));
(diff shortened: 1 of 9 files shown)


IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -23,43 +23,105 @@
...
+
+- (instancetype)initWithAcquirer:(id<MSIDSPATokenAcquiring>)acquirer
+{
+    self = [super init];
+    if (self)
+    {
+        _acquirer = acquirer;
+    }
+    return self;
+}
```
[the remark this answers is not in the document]
agubuzomaximus: Addressed in 5b6b6c51a. `initWithAcquirer:` now asserts the nonnull dependency immediately, preventing construction with an invalid acquirer.

IdentityCore/tests/MSIDBoundTokenProviderTests.m
```diff
@@ -25,128 +25,285 @@
...
+    outcome.fallbackRequestAccountUpn = account.username;
+    return outcome;
+}
+
+- (NSDictionary *)payloadFromResponse:(NSString *)response
 {
     NSData *data = [response dataUsingEncoding:NSUTF8StringEncoding];
-    XCTAssertNotNil(data);
+    return [NSJSONSerialization JSONObjectWithData:data options:0 error:nil];
+}
```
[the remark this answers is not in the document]
agubuzomaximus: Addressed in 5b6b6c51a. The helper now checks the response/data, captures and reports JSON parsing errors, validates the parsed object is a dictionary, and returns nil for invalid payloads.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +149,133 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+    {
+        NSString *routeReason = route == MSIDBrowserNativeMessageGetTokenRouteUIBlocked ? @"canShowUI is NO" : @"prompt=none forbids UI";
+        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
+                          @"%@ Interactive fallback rejected. Route: %ld. Reason: %@. promptType: %ld, canShowUI: %@.",
+                          MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, (long)route, routeReason, (long)parameters.promptType, request.canShowUI ? @"YES" : @"NO");
+        NSError *error = interactionRequiredError
+            ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
+                               @"Bound token acquisition requires user interaction.",
+                               nil, nil, nil, context.correlationId, nil, YES);
+        completionBlock(nil, error);
```
josephpab, line 215: **Severity: Medium** — `UIBlocked` and `InteractionRequired` collapse into the same error, so the host cannot distinguish them.

**Issue:** `MSIDBrowserNativeMessageGetTokenRoutingPolicy` deliberately returns two distinct non-interactive routes, but this branch maps both onto the same `MSIDErrorDomain` / `MSIDErrorInteractionRequired` error. `route` itself is only ever consumed by the log line. On the silent→interactive fallback path `interactionRequiredError` is non-nil, so a `canShowUI == NO` request is answered with the *silent* error — for the local acquirer that is `MSIDSilentTokenRequest`'s `"No token matching arguments found in the cache, user interaction is required"`, which says nothing about UI being blocked.

**Impact:** `MSIDBrowserNativeMessageGetTokenRequest.canShowUI` is documented as *"When this flag is 'false', ui is not allowed. If we cannot acquire tokens silently, UI_NOT_ALLOWED status should be returned to the browser."* With identical domain, code, and description for both routes, the caller has no signal to emit `UI_NOT_ALLOWED` rather than a generic interaction-required, and the routing policy's distinction is discarded at the only place it is consumed.

**Recommendation:** Keep the route observable to the caller and stop reusing the silent error's description for a UI-policy decision. Without inventing a new error code, `additionalUserInfo` plus a route-accurate description and the silent error chained as the underlying error is enough:

```objc
    if (route != MSIDBrowserNativeMessageGetTokenRouteInteractive)
    {
        BOOL uiBlocked = route == MSIDBrowserNativeMessageGetTokenRouteUIBlocked;
        NSString *routeReason = uiBlocked ? @"canShowUI is NO" : @"prompt=none forbids UI";
        MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
                          @"%@ Interactive fallback rejected. Route: %ld. Reason: %@. promptType: %ld, canShowUI: %@.",
                          MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, (long)route, routeReason, (long)parameters.promptType, request.canShowUI ? @"YES" : @"NO");
        NSString *description = uiBlocked
            ? @"Bound token acquisition requires user interaction, but the caller cannot show UI."
            : @"Bound token acquisition requires user interaction, but prompt=none forbids UI.";
        NSError *error = MSIDCreateError(MSIDErrorDomain, MSIDErrorInteractionRequired,
                                         description,
                                         interactionRequiredError.msidOauthError,
                                         interactionRequiredError.msidSubError,
                                         interactionRequiredError,
                                         context.correlationId,
                                         @{@"MSIDGetTokenRoute": @(route)}, YES);
        completionBlock(nil, error);
        return;
    }
```

(If the `UI_NOT_ALLOWED` mapping is intended to live in the OneAuth host and it already keys off `canShowUI` itself, say so and I'll drop this — but then the `UIBlocked` route serves no purpose here and the reused silent description is still misleading.)

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +149,133 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
+    [self.acquirer acquireInteractiveWithParameters:parameters
+                                            request:request
+                                            context:context
+                                    completionBlock:^(MSIDSPATokenAcquisitionResult *outcome, NSError *error) {
+        [self completeWithOutcome:outcome
+                            error:error
+                          request:request
+                          context:context
+                  completionBlock:completionBlock];
+    }];
```
josephpab, line 228: **Severity: Medium** — `interactionRequiredError` is dropped on the interactive path, losing the reason silent failed.

**Issue:** When the silent attempt returns interaction-required and routing then selects `Interactive`, `interactionRequiredError` is never used again. `MSIDSilentTokenRequest` wraps the real cause inside that error — see `MSIDSilentTokenRequest.m` L397, which passes `error.msidOauthError`, `error.msidSubError`, and the original `NSError` as the underlying error. So a revoked grant, an expired BART, a `invalid_grant`/sub-error pair, etc. all live only inside the error being discarded here.

**Impact:** If `acquireInteractiveWithParameters:` fails, the caller gets that backend's error with no trace of why silent failed. Today that is guaranteed, not hypothetical: `MSIDLocalSPATokenAcquirer` always completes with `"Interactive broker acquisition is required but not implemented."`, so every fallback in this release surfaces a message that discards the actual authentication failure. That will make the first round of field diagnostics on this flow considerably harder.

**Recommendation:** Chain the silent error through when the interactive attempt itself fails, e.g.:

```objc
    [self.acquirer acquireInteractiveWithParameters:parameters
                                            request:request
                                            context:context
                                    completionBlock:^(MSIDSPATokenAcquisitionResult *outcome, NSError *error) {
        NSError *resolvedError = error;
        if (!outcome && error && interactionRequiredError)
        {
            resolvedError = MSIDCreateError(error.domain, error.code, error.msidErrorDescription,
                                            error.msidOauthError, error.msidSubError,
                                            interactionRequiredError, context.correlationId, nil, YES);
        }

        [self completeWithOutcome:outcome
                            error:resolvedError
                          request:request
                          context:context
                  completionBlock:completionBlock];
    }];
```

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -86,28 +149,133 @@ - (BOOL)validateRequest:(MSIDBrowserNativeMessageGetTokenRequest *)request
...
-        return @"{}";
+        NSError *responseError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
+                                                 @"Failed to serialize bound token response.",
+                                                 nil, nil, serializationError, context.correlationId, nil, NO);
+        completionBlock(nil, responseError);
+        return;
     }
 
-    return [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
+    completionBlock([[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding], nil);
```
josephpab, line 278: **Severity: Medium** — every failure exit in this method is silent, and the success exit can hand back `(nil, nil)`.

**Issue (a) — no telemetry on failure.** All three failure exits build their error with `logErrorDescription: NO` (the final `MSIDCreateError` argument) and there is no `MSID_LOG_WITH_CTX` anywhere in this method. Compare with the parameters-mapping failure in `acquireBoundTokenWithRequest:`, which logs explicitly before returning. A response-shaping or serialization failure in production therefore emits nothing at all.

**Issue (b) — two distinct failures share one message.** `-[MSIDBrowserNativeMessageGetTokenResponse initWithTokenResult:state:fallbackRequestAccountUpn:]` returns `nil` when the result has neither an access token nor a token response. That `nil` flows into `[response jsonDictionary]`, which returns `nil` by message-to-nil, and lands in the same `"Failed to create bound token response."` branch as a genuine `jsonDictionary` failure. In the field these two are indistinguishable, and (a) means there is no log to tell them apart either.

**Issue (c) — `(nil, nil)` is reachable.** `-[NSString initWithData:encoding:]` is nullable. If it returns `nil`, the completion is invoked with no response *and* no error. This method already guards exactly that invariant at the top (`"Bound token acquisition completed without a result or error."`), so the last line contradicts the contract the rest of the method enforces.

**Impact:** A caller receiving `(nil, nil)` has no way to report or retry, and the missing logs remove the only diagnostics for the terminal step of the flow.

**Recommendation:**

```objc
    if (!outcome)
    {
        NSError *completionError = error
            ?: MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
                               @"Bound token acquisition completed without a result or error.",
                               nil, nil, nil, context.correlationId, nil, YES);
        MSID_LOG_WITH_CTX(MSIDLogLevelError, context, @"%@ Bound token acquisition failed: %@",
                          MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX, MSID_PII_LOG_MASKABLE(completionError));
        completionBlock(nil, completionError);
        return;
    }
```

…and likewise pass `YES` for `logErrorDescription` on the two response-shaping errors, split the `!response` case from the `!responseDictionary` case so the messages differ, and guard the final string:

```objc
    NSString *responseString = [[NSString alloc] initWithData:data encoding:NSUTF8StringEncoding];
    if (!responseString)
    {
        NSError *responseError = MSIDCreateError(MSIDErrorDomain, MSIDErrorInternal,
                                                 @"Failed to encode bound token response.",
                                                 nil, nil, nil, context.correlationId, nil, YES);
        completionBlock(nil, responseError);
        return;
    }

    completionBlock(responseString, nil);
```

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -23,43 +23,106 @@
...
+    }
 
-    MSID_LOG_WITH_CTX(MSIDLogLevelInfo, context,
-                      @"%@ In-process GetToken request completed.", MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX);
+    MSIDBrowserNativeMessageGetTokenRoute route =
+    [[self routingPolicy] routeWithForceInteractive:NO
+                                         promptType:parameters.promptType
+                                          canShowUI:request.canShowUI
+                                  accountIdentifier:parameters.accountIdentifier
+                              requiresHomeAccountId:NO];
```
josephpab, line 97: Non-blocking question — a `loginHint`-only GetToken request never attempts the in-process cache.

`MSIDInteractiveTokenRequestParameters+BrowserNativeMessageGetToken` populates `accountIdentifier` from `request.accountId` only (`parameters.accountIdentifier = request.accountId;`), and `-shouldAttemptSilentWithForceInteractive:...` returns `NO` outright when `accountIdentifier` is `nil`, independently of `requiresHomeAccountId`. So a request carrying only a `loginHint` routes straight to `Interactive` here, and with today's `MSIDLocalSPATokenAcquirer` that means an immediate interaction-required — the access-token cache is never consulted.

What makes me want to confirm rather than assume: the routing policy's own comment for the `requiresHomeAccountId:NO` case you're passing says *"STS and in-process BART callers can resolve an account from a displayable ID"*, which reads as though a displayable-ID-only request is meant to remain silent-eligible. But `nil` `accountIdentifier` short-circuits before that logic is reached.

Is `accountId` guaranteed to be present for the unmanaged BART SPA path? If yes, disregard. If not, consider synthesising an identifier from the hint when `accountId` is absent, so the displayable-ID case actually reaches the cache:

```objc
    MSIDAccountIdentifier *accountIdentifier = parameters.accountIdentifier;
    if (!accountIdentifier && ![NSString msidIsStringNilOrBlank:request.loginHint])
    {
        accountIdentifier = [[MSIDAccountIdentifier alloc] initWithDisplayableId:request.loginHint homeAccountId:nil];
    }
```

Either way it would be worth a test that pins the intended behaviour for `accountId == nil, loginHint != nil` — right now no test covers it (`validRequest` always sets `accountId`).

IdentityCore/src/oauth2/token/MSIDSPATokenAcquiring.h
```diff
@@ -0,0 +1,46 @@
...
+                                                                          error:(NSError *_Nullable *_Nullable)error;
+- (void)acquireSilentWithParameters:(MSIDInteractiveTokenRequestParameters *)parameters
+                            request:(MSIDBrowserNativeMessageGetTokenRequest *)request
+                            context:(nullable id<MSIDRequestContext>)context
+                    completionBlock:(MSIDSPATokenAcquirerCompletionBlock)completionBlock;
+- (void)acquireInteractiveWithParameters:(MSIDInteractiveTokenRequestParameters *)parameters
+                                request:(MSIDBrowserNativeMessageGetTokenRequest *)request
+                                context:(nullable id<MSIDRequestContext>)context
+                        completionBlock:(MSIDSPATokenAcquirerCompletionBlock)completionBlock;
+@end
```
josephpab, line 45: Non-blocking — this protocol is the extension seam for the broker-flip acquirer, but it carries no contract documentation.

`MSIDBoundTokenProvider`'s correctness is fully delegated to implementers of this protocol. I walked every path in `acquireBoundTokenWithRequest:` and the completion block is invoked exactly once on each — **but only if** `acquireSilentWithParameters:` and `acquireInteractiveWithParameters:` each invoke their own completion exactly once. An implementation that drops the callback hangs the browser request forever; one that double-invokes produces a double callback into the host. Nothing here states that requirement, and the two obvious future implementers (a broker-flip acquirer, and any OneAuth-side test double) will be written against this header alone.

Worth documenting on the protocol:
1. `completionBlock` **must** be invoked exactly once on every path, including early error returns.
2. Which queue the completion is delivered on. Today `MSIDLocalSPATokenAcquirer` completes synchronously on the caller's thread, and `MSIDBoundTokenProvider` also completes synchronously for validation/mapping errors — so the provider's public completion is sometimes sync and sometimes async. A real broker flip makes it always async. Callers should be told not to rely on either.
3. `requestParametersForRequest:context:error:` must populate `*error` whenever it returns `nil` — `MSIDBoundTokenProvider` wraps `parametersError` as the underlying error and would otherwise produce an error with an empty cause.

Minor readability note while you're here: this header has no blank lines between the imports, the typedef, and the protocol members, which reads differently from `MSIDLocalSPATokenAcquirer.h` and `MSIDBoundTokenProvider.h` in the same folder.

IdentityCore/src/oauth2/token/MSIDLocalSPATokenAcquirer.h
```diff
@@ -37,7 +38,7 @@ typedef void (^MSIDLocalSPATokenAcquirerCompletionBlock)(MSIDSPATokenAcquisition
 typedef MSIDDefaultSilentTokenRequest *_Nullable (^MSIDLocalSPASilentTokenRequestProvider)(
     MSIDInteractiveTokenRequestParameters *parameters, id<MSIDRequestContext> _Nullable context);
 
-@interface MSIDLocalSPATokenAcquirer : NSObject
+@interface MSIDLocalSPATokenAcquirer : NSObject <MSIDSPATokenAcquiring>
```
josephpab, line 41: Non-blocking — now that this class adopts `MSIDSPATokenAcquiring`, `MSIDLocalSPATokenAcquirerCompletionBlock` (L36) is a duplicate of `MSIDSPATokenAcquirerCompletionBlock`; both are `void (^)(MSIDSPATokenAcquisitionResult *_Nullable, NSError *_Nullable)`.

The `.m` already mixes them: `acquireSilentWithParameters:...` is declared with the local typedef while the new `acquireInteractiveWithParameters:...` uses the protocol's. It compiles because the underlying block types are identical, but two names for one contract is exactly the kind of thing that drifts the next time the acquirer signature changes — and a change to only one of them would fail at the conformance site rather than at the definition.

Suggest deleting `MSIDLocalSPATokenAcquirerCompletionBlock` and using `MSIDSPATokenAcquirerCompletionBlock` in both the header declaration and the implementation.

IdentityCore/tests/MSIDLocalSPATokenAcquirerTests.m
```diff
@@ -125,7 +125,7 @@ - (void)testAcquireSilent_whenEngineReturnsResult_shouldRequireBoundRefreshToken
 
     XCTAssertNil(acquisitionError);
     XCTAssertEqual(outcome.tokenResult, result);
-    XCTAssertEqualObjects(outcome.fallbackRequestAccountUpn, @"cached@contoso.com");
+    XCTAssertEqualObjects(outcome.fallbackRequestAccountUpn, @"user@contoso.com");
```
josephpab, line 128: **Severity: Medium** — repurposing this test removed the only coverage of the populated-username branch.

Before this PR the test set `result.account.username = @"cached@contoso.com"` and asserted `fallbackRequestAccountUpn == @"cached@contoso.com"`. It has been rewritten in place to cover the blank-username case, so the new ternary in `MSIDLocalSPATokenAcquirer.m` L109 —

```objc
outcome.fallbackRequestAccountUpn = [NSString msidIsStringNilOrBlank:username] ? request.loginHint : username;
```

— now has only its `msidIsStringNilOrBlank == YES` branch under test. The `NO` branch (cached username wins over the login hint) is unverified, and that is the branch that carries the actual account identity into `MSIDBrowserNativeMessageGetTokenResponse`'s `account.userName` / `properties.UPN`.

Given the whole point of this change is the blank-vs-populated distinction, both sides should be pinned. Recommend keeping this test as the blank case and re-adding the previous assertion as a separate test, e.g. `testAcquireSilent_whenEngineReturnsUsername_shouldPreferCachedUsernameOverLoginHint`, with `request.loginHint` deliberately set to a *different* value so the test fails if the ternary is ever inverted.

IdentityCore/tests/MSIDBoundTokenProviderTests.m
```diff
@@ -25,128 +25,285 @@
...
+#import "MSIDAccount.h"
+#import "MSIDAccountIdentifier.h"
+#import "MSIDAADAuthority.h"
 #import "MSIDError.h"
 
-@interface MSIDBoundTokenProviderTests : XCTestCase
+@interface MSIDSPATokenAcquirerMock : NSObject <MSIDSPATokenAcquiring>
+
+@property (nonatomic) BOOL silentCalled;
+@property (nonatomic) BOOL interactiveCalled;
```
josephpab, line 41: Non-blocking — `testAcquireBoundToken_whenSilentRequiresInteraction_shouldFallbackOnce` (L189) cannot actually observe "once".

`silentCalled` / `interactiveCalled` are booleans, so the test asserts only that each backend was invoked *at least* once. If a future change re-entered the routing loop — for example by re-running `routeWithForceInteractive:` on the interactive backend's own interaction-required error — this test would still pass while the flow spun.

Two suggestions:

1. Make these counters and assert the exact count, so the test name is backed by an assertion:
```objc
@property (nonatomic) NSUInteger silentCallCount;
@property (nonatomic) NSUInteger interactiveCallCount;
```
```objc
XCTAssertEqual(acquirer.silentCallCount, 1);
XCTAssertEqual(acquirer.interactiveCallCount, 1);
```

2. Add a companion case where `interactiveError` is *also* `MSIDErrorDomain` / `MSIDErrorInteractionRequired`. That is the realistic shape today, since `MSIDLocalSPATokenAcquirer.acquireInteractiveWithParameters:` always returns exactly that error. `completeInteractiveRoute:` → `completeWithOutcome:` is terminal, so the correct expectation is `interactiveCallCount == 1` and the interactive error propagated verbatim — worth locking in explicitly rather than leaving it implied.

IdentityCore/src/oauth2/token/MSIDLocalSPATokenAcquirer.h
jasoncoolmax, line 39: We now have two typedefs with the identical signature, `MSIDLocalSPATokenAcquirerCompletionBlock` here and `MSIDSPATokenAcquirerCompletionBlock` in the protocol, and `acquireSilentWithParameters:` below is still declared with the old one. Shall we drop the local typedef and use the protocol's everywhere? :)

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: The `UI_NOT_ALLOWED` browser status is intentionally mapped by the OneAuth host, which owns the Browser Native Message response and still has the original `canShowUI` value. CommonCore uses the distinct route to prevent an interactive backend call; introducing a new NSError user-info contract here would duplicate the host mapping and create an unplanned cross-layer API. Keeping this unchanged.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: Keeping the interactive backend error as the terminal error is intentional. The silent failure and interactive failure are sequential attempts rather than a causal error chain, so replacing the interactive error underlying cause with the earlier silent error would discard any underlying error produced by the interactive backend. Silent BART failures are logged at their source; the caller receives the actionable terminal failure.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: Addressed in c69bcbc0c. Each locally-created terminal error now logs through `MSIDCreateError`, response initialization and dictionary shaping have distinct descriptions, and UTF-8 conversion is guarded so completion cannot return `(nil, nil)`.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: `MSIDBrowserNativeMessageGetTokenRequest` already synthesizes `accountId` from `loginHint` during JSON parsing (`initWithDisplayableId:homeAccountId:` whenever either value is present). The unmanaged Browser Native Message path therefore reaches routing with a displayable-ID account identifier even when the JSON omits the home account ID. Adding a second synthesis in the provider would duplicate request parsing behavior.

IdentityCore/src/oauth2/token/MSIDSPATokenAcquiring.h
agubuzomaximus: Addressed in c69bcbc0c. The protocol now documents exactly-once completion delivery, its synchronous/asynchronous arbitrary-queue behavior, and the error requirement when parameter mapping returns nil.

IdentityCore/src/oauth2/token/MSIDLocalSPATokenAcquirer.h
agubuzomaximus: Addressed in c69bcbc0c. Removed the duplicate local completion typedef and use `MSIDSPATokenAcquirerCompletionBlock` consistently in the declaration and implementation.

IdentityCore/tests/MSIDLocalSPATokenAcquirerTests.m
agubuzomaximus: Addressed in c69bcbc0c. The test now exercises both branches with different login-hint and cached usernames, verifying blank usernames fall back to the hint and populated cached usernames win.

IdentityCore/tests/MSIDBoundTokenProviderTests.m
agubuzomaximus: Addressed the exact-once concern in c69bcbc0c by replacing booleans with call counters and asserting one silent and one interactive invocation. I did not add the suggested duplicate interactive-error case because interactive completion is terminally passed to `completeWithOutcome:` and is never routed again, so it would not exercise a separate routing branch.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: You were right. I compared this with Broker production behavior: `ADBBrowserNativeMessageGetTokenOperation` attaches `MSALBrowserNativeMessageErrorStatus = UI_NOT_ALLOWED` when interaction is permitted but `canShowUI` is false, while prompt-none maps to `USER_INTERACTION_REQUIRED`. Addressed in 192cd0ade by preserving those statuses and route-specific descriptions in CommonCore, including the silent-fallback path.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
agubuzomaximus: I rechecked this against Broker production and am keeping the current behavior. After a silent interaction-required result, `ADBBrowserNativeMessageGetTokenOperation` restarts with `forceInteractive = YES`; if the interactive operation fails, it returns that interactive error and does not chain the earlier silent error. CommonCore matches that behavior because these are sequential attempts, not a causal error chain, and preserving the interactive backend error also preserves its own underlying cause.

IdentityCore/src/oauth2/token/MSIDLocalSPATokenAcquirer.h
agubuzomaximus: Yes—addressed in c69bcbc0c. The duplicate `MSIDLocalSPATokenAcquirerCompletionBlock` typedef was removed, and both the header and implementation now use `MSIDSPATokenAcquirerCompletionBlock` from the protocol.

antonioalwan (approved):

antonioalwan (approved):

PR #1915, opened 2026-07-21
[major] [engg]: Bump minimum deployment targets

josephpab:
## PR Checklist (must be completed before review)

- [x] All tests pass locally
- [x] PR size is <= 500 LOC per PR Size Check policy
- [x] PR is independently mergeable (no hidden dependencies)
- [x] Appropriate reviewers are assigned
- [x] PR reviewed by code owner (required if Copilot-generated)
- [x] SME or Senior IC assigned where required

## PR Title Format

**Required Format:** `[Keyword1] [Keyword2]: Description`

- **Keyword1:** `major`, `minor`, or `patch` (case-insensitive)
- **Keyword2:** `feature`, `bugfix`, `engg`, or `tests` (case-insensitive)

**Examples:**
- `[MAJOR] [Feature]: new API`
- `[minor] [bugfix]: fix crash`
- `[PATCH][tests]:add coverage`

## Proposed changes

Raise all IdentityCore minimum deployment targets to iOS 16.0 and macOS 12.0 across the shared xcconfigs and Xcode project configurations.

Replace the deprecated macOS `SecTransform` RSA SHA-256 digest-signing path in `NSData+JWT` with `SecKeyCreateSignature`, preserving the existing PKCS#1 v1.5 signing behavior.

## Type of change

- [ ] Feature work
- [ ] Bug fix
- [ ] Documentation
- [x] Engineering change
- [ ] Test
- [ ] Logging/Telemetry

## Risk

- [ ] High – Errors could cause MAJOR regression of many scenarios. (Example: new large features or high level infrastructure changes)
- [x] Medium – Errors could cause regression of 1 or more scenarios. (Example: somewhat complex bug fixes, small new features)
- [ ] Small – No issues are expected. (Example: Very small bug fixes, string changes, or configuration settings changes)

## Additional information

Validation completed:

- Targeted macOS RSA signature equivalence test
- IdentityCore Mac Release build
- IdentityCore iOS Release build for the iOS Simulator

PR #1934, opened 2026-08-10
[minor] [feature]: Add external key support for AT PoP

antonioalwan:
## Summary

Adds caller-provided RSA key support for Access Token Proof-of-Possession in CommonCore. The change validates external key pairs, supports signing through an external `MSIDDevicePopManager`, preserves external-key classification across Broker request and resume paths, emits low-cardinality key-source telemetry, and verifies cache isolation by key ID.

## Key changes

- Validate caller-provided RSA key handles, size, capabilities, and pair matching.
- Allow `MSIDDevicePopManager` to use a caller-provided asymmetric key pair.
- Preserve `external_key_pop=1` through auth-scheme JSON, Broker query, and resume state.
- Emit `pop_key_source=internal|external` telemetry without key material.
- Add validation, signing, marker-preservation, Broker serialization, and cache-isolation tests.

## How to validate

- Run `MSIDAuthenticationSchemePopTest`.
- Run `MSIDBrokerTokenRequestTests`.
- Run `MSIDCredentialCacheItemTests`.
- Run the external key validation and device PoP manager tests.

## Notes

This is the first PR in the CommonCore -> MSAL -> Broker dependency chain.


IdentityCore/src/pop_manager/MSIDDevicePopManager.m
```diff
@@ -56,6 +56,17 @@ - (instancetype)initWithCacheConfig:(MSIDCacheConfig *)cacheConfig
     return self;
 }
 
+- (instancetype)initWithExternalKeyPair:(MSIDAssymetricKeyPair *)keyPair
```
kaisong1990, line 59: This accepts nil without guarding. If keyPair is nil, the lazy `keyPair` getter below still runs and calls `readOrGenerateKeyPairForAttributes:` on a nil keyGeneratorFactory with nil attributes, so we silently get nil back and every signing call fails later with `unable to generate kid`. Please return nil here when keyPair is nil and mark the initializer nullable in the header.

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,164 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+                                                    error:error];
+            }
+
+            return YES;
+        }
+
+        if (derivedKeyError) CFRelease(derivedKeyError);
+    }
+
+    NSData *challenge = [@"MSAL external AT PoP key pair validation" dataUsingEncoding:NSUTF8StringEncoding];
```
kaisong1990, line 238: This fallback signs with the caller's private key. If the app created that key with an access control that requires user presence (kSecAccessControlUserPresence / biometry), SecKeyCreateSignature will put up a LocalAuthentication prompt and block the calling thread. Validation is invoked synchronously from MSALExternalKeyPair init, so an app constructing the key pair on the main thread gets a surprise Face ID sheet. Since we already have the public key handle from the caller, we don't need to sign to prove the pair matches. Consider dropping this path and failing validation when SecKeyCopyPublicKey / serialization doesn't give us something to compare.

IdentityCore/src/util/MSIDKeyOperationUtil.h
```diff
@@ -68,5 +81,10 @@ NS_ASSUME_NONNULL_BEGIN
                                 signingAlgorithm:(SecKeyAlgorithm)algorithm
                                          context:(_Nullable id<MSIDRequestContext>)context
                                            error:(NSError * _Nullable __autoreleasing * _Nullable)error;
+
+- (BOOL)validateExternalRSAKeyPair:(SecKeyRef _Nullable)privateKey
```
kaisong1990, line 85: Every other public method on this class takes a `context:` so failures can be correlated in logs. This one doesn't, and `failExternalKeyPairValidation:` ends up passing `context:nil` to `generateErrorWithMessage:`. Please add a context parameter and thread it through.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
antonioalwan: Fixed in 409867ef9: the initializer is now nullable, rejects a nil key pair immediately, and has regression coverage.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: Fixed in 409867ef9: validation no longer signs with the caller-provided private key. It derives and serializes the public key for comparison, avoiding user-presence/biometric prompts.

IdentityCore/src/util/MSIDKeyOperationUtil.h
antonioalwan: Fixed in 409867ef9: the validation API now accepts and propagates MSIDRequestContext, with a test verifying the correlation ID reaches the generated error.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -28,7 +28,7 @@
 #import "MSIDLogger+Internal.h"
 #import "NSString+MSIDExtensions.h"
 
-NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
+static NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
```
ameyapat, line 37: Is this change related to the AT-PoP feature?

josephpab (commented):
Thorough pass over this one. The plumbing (marker propagation through auth-scheme JSON, broker query, resume state, telemetry) is clean and consistent, and cache isolation via `kid` is sound since `MSIDCredentialCacheItem` already includes `kid` in both `isEqualToItem:` and `hash`. Comments below are ordered roughly by severity.

**Main concerns**

1. **The validation is dead code.** `validateExternalRSAKeyPair:...` has no production caller — only tests. `initWithExternalKeyPair:` accepts any key pair unchecked. As it stands, CommonCore enforces nothing; enforcement is deferred to a downstream PR that has not merged.
2. **Validation probes a different `SecKeyAlgorithm` than the signing path gates on.** Validation checks the `Message` PKCS1v15 SHA-256 variant; `getJwtAlgorithmForKey:` (which `createSignedAccessToken` requires to succeed) checks the `Digest` variant. Token/smartcard-backed RSA keys frequently support only one of the two, so a key can pass validation and still fail at signing.
3. **`external_key_pop` reaches broker with no feature flag or version negotiation.** An older broker ignores the unknown parameter and returns a token bound to its own key while the app assumes the caller key — a silent binding mismatch rather than a clean failure. The adjacent `MSID_BOUND_RT_REDEEM` parameter is gated behind `MSIDBartFeatureUtil`; this one is not.

**Smaller items**

4. The 409867e auto-fix removed the sign/verify fallback, narrowing support for keys where `SecKeyCopyPublicKey` returns NULL — worth documenting as a deliberate tradeoff.
5. `2048` as a magic number, and a missing `kSecAttrKeySizeInBits` being reported as `KeySizeTooSmall`.
6. `MSIDBaseRequestController` re-derives the marker from the raw dictionary instead of using the scheme's already-parsed value, with a permissive `-boolValue`.
7. No test for the resume-state restore direction in `MSIDBrokerResponseHandler`.
8. The `MSIDBoundTokenProvider` `static` fix is unrelated scope and has no changelog entry.
9. Changelog entry is missing the `(#1934)` reference.

**On the tests** — `testMatchAccessTokenKeyThumbprint_whenKeysDiffer_shouldIsolateCachedTokens` and the `MSIDCredentialCacheItem` tests are valid (I verified the `req_cnf` values decode to distinct `kid`s, and that `kid` participates in equality/hash), but they are regression guards on pre-existing behaviour rather than coverage of anything this PR changes. The coverage gap that matters is item 7.

Items 1–3 look blocking to me; the rest are straightforward to address.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
```diff
@@ -56,7 +56,23 @@ - (instancetype)initWithCacheConfig:(MSIDCacheConfig *)cacheConfig
...
+{
+    if (!keyPair)
+    {
+        return nil;
+    }
+
+    self = [super init];
+    if (self)
+    {
+        _keyPair = keyPair;
```
josephpab: **The validation this PR adds is never actually enforced.**

`validateExternalRSAKeyPair:publicKey:failureReason:context:error:` has zero production callers — `git grep` finds it only in `MSIDKeyOperationUtilTest.m`. This initializer accepts any `MSIDAssymetricKeyPair` and installs it directly as the signing key with no checks at all.

So the PR summary's "Validate caller-provided RSA key handles, size, capabilities, and pair matching" is only true if the downstream MSAL PR remembers to call the validator. As written, a 1024-bit key, an EC key, or a public-key-only handle flows straight through to `SecKeyCreateSignature` in `createSignedAccessToken`.

Please close the loop inside CommonCore instead of relying on the caller — e.g. `initWithExternalKeyPair:context:error:` that validates and returns `nil` with a populated `NSError` on failure. Enforcing at the boundary means the invariant holds for OneAuth and Broker too, not just the one MSAL call site being written next. If the split really must stay, at minimum document the precondition in `MSIDDevicePopManager.h` so it is discoverable.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
```diff
@@ -56,7 +56,23 @@ - (instancetype)initWithCacheConfig:(MSIDCacheConfig *)cacheConfig
     return self;
 }
 
- - (MSIDAssymetricKeyPair *)keyPair
+- (instancetype)initWithExternalKeyPair:(MSIDAssymetricKeyPair *)keyPair
```
josephpab, line 60: The external-key contract here rests on an accident rather than on the type.

The `keyPair` getter just below still does lazy generation whenever `_keyPair` is nil:

```objc
- (MSIDAssymetricKeyPair *)keyPair
{
    if (!_keyPair)
    {
        _keyPair = [self.keyGeneratorFactory readOrGenerateKeyPairForAttributes:self.keyPairAttributes error:&keyPairError];
```

On this init path `cacheConfig`, `keyGeneratorFactory` and `keyPairAttributes` are all nil. It happens to be safe today only because `_keyPair` is guaranteed non-nil, but if it is ever cleared the getter messages nil, logs "Failed to generate key pair", and returns nil — a confusing failure for an object that by definition must never generate a key.

Suggest an explicit `_usesExternalKeyPair` flag with an early return in the getter, so "never generate" is a property of the instance rather than an emergent one. It also makes the intent obvious to the next reader.

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,162 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+    if (![privateKeySize isEqualToNumber:publicKeySize])
+    {
+        return [self failExternalKeyPairValidation:MSIDExternalKeyPairValidationFailureReasonKeyPairMismatch
+                                          message:@"External AT PoP RSA keys must have matching key sizes."
+                                    failureReason:failureReason
+                                          context:context
+                                            error:error];
+    }
+
+    SecKeyAlgorithm algorithm = kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256;
```
josephpab: **Validation probes a different algorithm than the signing path gates on, so passing validation does not guarantee signing will work.**

Here you check `kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256`. But `MSIDDevicePopManager.createSignedAccessToken` gates on `getJwtAlgorithmForKey:`, which probes the **Digest** variant:

```objc
if ([self isOperationSupportedByKey:kSecKeyOperationTypeSign
                          algorithm:kSecKeyAlgorithmRSASignatureDigestPKCS1v15SHA256 ...])
{
    return MSID_JWT_ALG_RS256;
}
```

and bails with "Key signing algorithm not supported." when that returns nil. `Message` and `Digest` are distinct `SecKeyAlgorithm` values and `SecKeyIsAlgorithmSupported` can legitimately disagree between them — CryptoTokenKit/smartcard-backed RSA keys commonly advertise only the `Digest` variants because they cannot hash on-token. That is precisely the caller-provided-key scenario this feature exists for.

Net effect: a caller key can pass validation and still fail at signing time. Please validate using the same resolution the signing path uses (`getJwtAlgorithmForKey:`), or probe both variants.

Related: this function hard-rejects EC keys with `UnsupportedKeyType`, while `getJwtAlgorithmForKey:` and `MSIDJWTHelper` fully support ES256 and would happily sign with one. If RSA-only is a deliberate service-side constraint that is fine, but right now the restriction lives exclusively in a function nothing calls — so the actual runtime behaviour is "EC works".

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,162 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+        return [self failExternalKeyPairValidation:MSIDExternalKeyPairValidationFailureReasonInvalidKeyClass
+                                          message:@"External AT PoP keys must contain one private key and one public key."
+                                    failureReason:failureReason
+                                          context:context
+                                            error:error];
+    }
+
+    NSNumber *privateKeySize = privateAttributes[(id)kSecAttrKeySizeInBits];
+    NSNumber *publicKeySize = publicAttributes[(id)kSecAttrKeySizeInBits];
+    if (privateKeySize.integerValue < 2048 || publicKeySize.integerValue < 2048)
```
josephpab: Two issues in this size check.

1. `2048` is a magic number repeated twice. Please hoist it to a named constant (`kMSIDMinimumExternalPoPKeySizeInBits`) so the policy is greppable and changeable in one place.

2. `kSecAttrKeySizeInBits` is not guaranteed to be present in every `SecKeyCopyAttributes` result, particularly for token/HSM-backed keys. When it is absent, `privateKeySize` is nil, `.integerValue` is `0`, and the caller is told `KeySizeTooSmall` for a key whose size simply could not be read. That is a misleading diagnostic and it will pollute the failure-reason telemetry this PR is adding.

Prefer an explicit nil check reported as `InvalidKeyHandle` / "unable to read key attributes" before the numeric comparison.

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,162 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+        NSError *underlyingError = CFBridgingRelease(publicKeyError);
+        return [self failExternalKeyPairValidation:MSIDExternalKeyPairValidationFailureReasonPublicKeySerializationFailed
+                                          message:@"External AT PoP public key cannot be serialized."
+                                 underlyingError:underlyingError
+                                    failureReason:failureReason
+                                          context:context
+                                            error:error];
+    }
+
+    SecKeyRef derivedPublicKey = SecKeyCopyPublicKey(privateKey);
```
josephpab, line 281: Commit 409867e turned the `SecKeyCopyPublicKey` fallback into a hard failure, and that is a behaviour reduction worth calling out explicitly rather than landing as cleanup.

The original implementation fell back to a sign-then-verify challenge when the public key could not be derived from the private key. `SecKeyCopyPublicKey` returns NULL for keys whose provider cannot hand back a public key handle — again, token/HSM-backed keys, the primary external-key scenario. Those key pairs are now rejected outright.

If the fallback was removed because `SecKeyCreateSignature` can trigger a biometric/PIN prompt during what the caller thinks is a cheap validation call, that is a perfectly good reason — but please state it in a code comment so it does not get "helpfully" restored later, and consider whether it belongs in the changelog as a documented limitation.

Separately: reporting `PublicKeySerializationFailed` for a *derivation* failure conflates two different conditions in the enum that exists specifically to distinguish them. Worth a dedicated case (`PublicKeyDerivationFailed`).

IdentityCore/src/requests/broker/MSIDBrokerTokenRequest.m
```diff
@@ -180,6 +181,7 @@ - (NSDictionary *)defaultPayloadContents:(NSError *__autoreleasing*)error
     [queryDictionary msidSetNonEmptyString:self.brokerApplicationToken forKey:@"application_token"];
     [queryDictionary msidSetNonEmptyString:tokenType forKey:MSID_OAUTH2_TOKEN_TYPE];
     [queryDictionary msidSetNonEmptyString:requestConf forKey:MSID_OAUTH2_REQUEST_CONFIRMATION];
+    [queryDictionary msidSetNonEmptyString:externalKeyPop forKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
```
josephpab, line 184: **This new broker parameter is not feature-gated or version-negotiated, and the failure mode is silent.**

`external_key_pop` is added to every PoP broker request unconditionally. Compare with the directly comparable new parameter ~15 lines above, which *is* gated:

```objc
if ([[MSIDBartFeatureUtil sharedInstance] isBartFeatureEnabled])
{
    [queryDictionary msidSetNonEmptyString:@"1" forKey:MSID_BOUND_RT_REDEEM];
}
```

What does a broker build that predates this parameter do? It will ignore the unknown key and mint a PoP token bound to the **broker's** key, while the app believes the token is bound to the caller-supplied key. That is a silent key-binding mismatch rather than a clean failure — the app will then sign PoP proofs with a key that has nothing to do with the `cnf` claim in the token, and failures surface as opaque 401s from the resource.

Please confirm there is a minimum broker version or an `sdkBrokerCapabilities` entry gating this. If there is not, this needs one — or CommonCore should refuse the broker route entirely when the external-key marker is set and the broker cannot honour it. Failing closed matters more than usual here because the consequence is a token bound to the wrong key.

IdentityCore/src/controllers/MSIDBaseRequestController.m
```diff
@@ -84,6 +86,14 @@ - (MSIDTelemetryAPIEvent *)telemetryAPIEvent
     [event setClientId:self.requestParameters.clientId];
     NSString *extExpiresSetting = self.requestParameters.extendedLifetimeEnabled ? MSID_TELEMETRY_VALUE_YES : MSID_TELEMETRY_VALUE_NO;
     [event setExtendedExpiresOnSetting:extExpiresSetting];
+    NSString *tokenType = self.requestParameters.authScheme.tokenType;
+    if (tokenType && [tokenType caseInsensitiveCompare:MSID_OAUTH2_POP] == NSOrderedSame)
+    {
+        BOOL isExternalKeyPop = [self.requestParameters.authScheme.schemeParameters[MSID_OAUTH2_EXTERNAL_KEY_POP] boolValue];
```
josephpab: Two sources of truth for the same fact, plus a lenient parse.

`MSIDAuthenticationSchemePop` already parses this marker into `_externalKeyPop` in its initializer, but that property is private to the `.m`, so this reaches back into the raw `schemeParameters` dictionary and re-derives it. Any future change to how the marker is represented has to be made in both places, and only one of them is covered by tests.

`-boolValue` is also looser than the wire contract: `@"yes"`, `@"true"` and `@"2"` all yield `YES`, while `@"0"` yields `NO`. And because `msidSetNonEmptyString:` silently drops non-`NSString` values on the `initWithJSONDictionary:` path, an `external_key_pop: 1` (JSON number) arriving from resume state degrades to `internal` with no log and no signal — the telemetry would then under-report external usage, which defeats the point of the metric.

Suggest exposing `- (BOOL)isExternalKeyPop` on the scheme (or on `MSIDAuthenticationScheme` returning `NO` by default) and calling that here, so parsing and interpretation live in exactly one tested place.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
```diff
@@ -201,8 +201,10 @@ - (MSIDAuthenticationScheme *)authSchemeFromResumeState:(NSDictionary *)resumeSt
     NSMutableDictionary *schemeParams = [NSMutableDictionary new];
     NSString *tokenType = resumeState[MSID_OAUTH2_TOKEN_TYPE];
     NSString *requestConf = resumeState[MSID_OAUTH2_REQUEST_CONFIRMATION];
+    NSString *externalKeyPop = resumeState[MSID_OAUTH2_EXTERNAL_KEY_POP];
     [schemeParams msidSetNonEmptyString:tokenType forKey:MSID_OAUTH2_TOKEN_TYPE];
     [schemeParams msidSetNonEmptyString:requestConf forKey:MSID_OAUTH2_REQUEST_CONFIRMATION];
+    [schemeParams msidSetNonEmptyString:externalKeyPop forKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
```
josephpab, line 208: This restore direction has no test coverage.

`MSIDBrokerTokenRequestTests` covers writing `external_key_pop` into the broker query and the resume dictionary, but nothing exercises reading it back out here and confirming the reconstructed `MSIDAuthenticationSchemePop` still carries the marker. Given that the whole point of the change is that the classification survives the app-relaunch round trip, that is the single most valuable assertion in the set and it is the one that is missing.

A small test that feeds a resume-state dictionary into `authSchemeFromResumeState:` and asserts `schemeParameters[MSID_OAUTH2_EXTERNAL_KEY_POP]` survives would cover it.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
josephpab, line 37: Unrelated to external AT PoP keys — this is a linker symbol-visibility fix that happens to be riding along in a feature PR.

It is a correct fix, but bundling it here means it cannot be cherry-picked independently if the feature slips, and it reviews under the wrong context. It also has no changelog entry, while the feature does — a consumer hitting the duplicate-symbol link failure between MSAL and OneAuth would have no way to find out it was fixed.

Please split it into its own PR, or at minimum add a changelog line for it.

changelog.txt
```diff
@@ -1,5 +1,6 @@
 TBD
 * Restrict macOS CBA preferred-identity URL fallback to the same host or its child subdomains behind a flight, with non-PII rollout logs.
```
josephpab, line 8: Nit: every other entry in this file references its PR (`(#1917)`, `(#1915)`, `(#1812)`), this one does not. Please add `(#1934)` for consistency and traceability.

Also, the entry describes the validator and the manager path as if they work together, but as noted in `MSIDDevicePopManager.m` the validator has no production caller in this PR. Worth wording it so it does not over-promise what CommonCore actually enforces today.

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,162 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+                          context:(id<MSIDRequestContext>)context
+                            error:(NSError *__autoreleasing *)error
+{
+    if (failureReason) *failureReason = MSIDExternalKeyPairValidationFailureReasonNone;
+
+    if (privateKey == NULL || publicKey == NULL)
+    {
+        return [self failExternalKeyPairValidation:MSIDExternalKeyPairValidationFailureReasonInvalidKeyHandle
+                                          message:@"Both private and public key handles are required."
+                                    failureReason:failureReason
```
ameyapat, line 155: Are we passing failureReason = None here? Is that right?

IdentityCore/src/util/MSIDKeyOperationUtil.m
ameyapat: Should the check here be to exactly equal 2048?

IdentityCore/src/util/MSIDKeyOperationUtil.m
ameyapat: nit: Maybe use existing isOperationSupportedByKey() in this class?

IdentityCore/src/util/MSIDKeyOperationUtil.m
```diff
@@ -135,6 +135,162 @@ - (NSData *)getSignatureForDataWithKey:(NSData *)rawData privateKey:(SecKeyRef)p
...
+        NSError *underlyingError = CFBridgingRelease(derivedKeyError);
+        return [self failExternalKeyPairValidation:MSIDExternalKeyPairValidationFailureReasonPublicKeySerializationFailed
+                                          message:@"The public key derived from the external AT PoP private key cannot be serialized."
+                                 underlyingError:underlyingError
+                                    failureReason:failureReason
+                                          context:context
+                                            error:error];
+    }
+
+    if (![derivedPublicKeyData isEqualToData:publicKeyData])
```
ameyapat, line 307: Why not just require caller to pass in only private key and we can derive the public key from it? Why ask caller to pass in public key and re-verify here if it is the same public key that is derived from private key?

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
antonioalwan: Agreed. I removed this unrelated linker change from #1934 and moved it, with its own changelog entry, to the focused PR #1937.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
antonioalwan: Addressed in 3bc9de3bc. `MSIDDevicePopManager` now validates caller-owned key pairs at its own initialization boundary through the context/error initializer, and the compatibility initializer delegates to it. The tests also verify that invalid pairs are rejected by the manager itself.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
antonioalwan: Addressed in 3bc9de3bc. The manager now records explicit external-key mode. If its caller-owned key is unexpectedly unavailable, `keyPair` fails closed instead of lazily generating a replacement. A regression test verifies that the key generator is not invoked.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: Addressed in 3bc9de3bc. Validation now requires both algorithms used by the runtime path: the RSA digest algorithm selected by `getJwtAlgorithmForKey:` and `kSecKeyAlgorithmRSASignatureMessagePKCS1v15SHA256`, which `MSIDJWTHelper` uses for JWT signing. RSA remains intentional because the public MSAL external-key API currently defines an RSA key-pair contract.

antonioalwan: Addressed in 3bc9de3bc. The capability check now calls `getJwtAlgorithmForKey:` and requires `RS256`, so validation and runtime algorithm selection use the same helper. It additionally verifies the message signing/verification algorithm used by `MSIDJWTHelper`.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: Addressed in 3bc9de3bc. The minimum is now the named `MSIDMinimumExternalPoPKeySizeInBits` constant, and missing/non-numeric key-size attributes are reported as invalid key handles rather than being conflated with an undersized key.

antonioalwan: I kept this as a minimum-size policy (`>= 2048`) rather than requiring exactly 2048 bits. Larger RSA keys satisfy the security requirement and are supported by Security.framework; rejecting them would unnecessarily narrow the public API contract. The named constant makes that policy explicit.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: Addressed in 3bc9de3bc and 84a43da2f. A derivation failure now has the dedicated `PublicKeyDerivationFailed` reason. The public initializer/validation documentation also states that derivation is required and that validation deliberately does not sign a challenge, because caller-owned Secure Enclave/HSM/token keys may trigger user-presence UI during initialization.

IdentityCore/src/requests/broker/MSIDBrokerTokenRequest.m
antonioalwan: I traced the legacy Broker path and do not believe this failure mode occurs. `ADBrokerRequest` copies the caller-provided `req_cnf` into the protocol scheme parameters, and `MSIDBrokerTokenRequest` forwards that same value to ESTS. A Broker that does not understand `external_key_pop` therefore still uses the caller key confirmation; it does not substitute a broker-generated key. The marker is used for ECS enforcement, telemetry, and round-trip classification, so no compatibility code change is needed here.

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
antonioalwan: Agreed. The bound-token symbol fix and its changelog entry were removed from #1934 and moved to the focused PR #1937, which can be reviewed and shipped independently.

IdentityCore/src/controllers/MSIDBaseRequestController.m
antonioalwan: Addressed in 3bc9de3bc. Marker parsing is centralized in `MSIDAuthenticationSchemePop`: only the string value `"1"` is accepted, malformed strings and numeric values are rejected, and controller telemetry now uses the typed `isExternalKeyPop` property instead of raw `boolValue` parsing.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
antonioalwan: Addressed in 3bc9de3bc. `MSIDBrokerResponseHandlerTests` now reconstructs a PoP scheme from resume state and verifies that `external_key_pop == "1"` survives and produces `isExternalKeyPop == YES`.

changelog.txt
antonioalwan: Addressed in 3bc9de3bc. The changelog entry now includes `(#1934)` and states that caller-provided key validation is enforced by the external-key manager initialization path.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: `MSIDExternalKeyPairValidationFailureReasonNone` is intentionally the success/default value. Validation initializes the out parameter to `None` and overwrites it on every failure path; when initialization succeeds, there is no failure reason to report.

IdentityCore/src/util/MSIDKeyOperationUtil.m
antonioalwan: Both handles are required for different purposes: the private key signs the AT PoP JWT, while the public key supplies the JWK/`kid` sent in `req_cnf`. Validation derives a public key from the private handle only to prove that the caller-provided public handle belongs to the same pair. This is now documented on the external-key initializer in 84a43da2f.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
```diff
@@ -56,10 +57,46 @@ - (instancetype)initWithCacheConfig:(MSIDCacheConfig *)cacheConfig
...
+{
+    MSIDExternalKeyPairValidationFailureReason validationReason = MSIDExternalKeyPairValidationFailureReasonNone;
+    BOOL valid = [[MSIDKeyOperationUtil sharedInstance] validateExternalRSAKeyPair:keyPair.privateKeyRef
+                                                                        publicKey:keyPair.publicKeyRef
+                                                                    failureReason:&validationReason
+                                                                          context:context
+                                                                            error:error];
+    if (!valid)
+    {
+        return nil;
```
kaisong1990, line 83: Returning nil here before `[super init]` leaks the allocation. In an init-family method ARC hands the method ownership of the receiver, and that reference is only balanced once `[super init]` runs, so this early exit drops it on the floor (clang's analyzer flags it). `test_initWithExternalKeyPair_NilKeyPair_ShouldReturnNil` hits this path directly.

Move `self = [super init];` above the validation and fail after it, which is also what the designated initializer above and `MSIDAuthenticationSchemePop` already do.

antonioalwan: Addressed in f476e3593. The initializer now calls `[super init]` before external-key validation, returns nil if superclass initialization fails, and performs the validation only after ARC has established the normal initializer ownership path.

IdentityCore/src/pop_manager/MSIDDevicePopManager.m
```diff
@@ -56,10 +57,48 @@ - (instancetype)initWithCacheConfig:(MSIDCacheConfig *)cacheConfig
...
+    _usesExternalKeyPair = YES;
+
+    return self;
+}
+
+- (MSIDAssymetricKeyPair *)keyPair
 {
     if (!_keyPair)
     {
+        if (self.usesExternalKeyPair)
```
jasoncoolmax, line 96: `initWithExternalKeyPair:` returns nil unless validation passes, and it always sets `_keyPair` on success, so `usesExternalKeyPair` can never be YES while `_keyPair` is nil. Seems this branch is unreachable, is it just defensive?

antonioalwan: Yes, under the current initializer contract this branch is unreachable.

`initWithExternalKeyPair:context:error:` either:
1. fails validation and returns nil, or
2. succeeds and sets both `_keyPair` and `_usesExternalKeyPair = YES`.

So with the current API surface, we should never observe `usesExternalKeyPair == YES` with a nil `_keyPair`.

IdentityCore/src/controllers/MSIDBaseRequestController.m
```diff
@@ -84,6 +87,15 @@ - (MSIDTelemetryAPIEvent *)telemetryAPIEvent
     [event setClientId:self.requestParameters.clientId];
     NSString *extExpiresSetting = self.requestParameters.extendedLifetimeEnabled ? MSID_TELEMETRY_VALUE_YES : MSID_TELEMETRY_VALUE_NO;
     [event setExtendedExpiresOnSetting:extExpiresSetting];
+    NSString *tokenType = self.requestParameters.authScheme.tokenType;
+    if (tokenType && [tokenType caseInsensitiveCompare:MSID_OAUTH2_POP] == NSOrderedSame)
+    {
+        BOOL isExternalKeyPop = [self.requestParameters.authScheme isKindOfClass:MSIDAuthenticationSchemePop.class]
```
antrix1989: Lets avoid dynamic casting. Can we add a method to the the base class that will populate event with needed fields instead? i.e: 
`[self.requestParameters.authScheme configureTelemetryEvent:event];`

By default this will do nothing, but for MSIDAuthenticationSchemePop this will set needed field.

antonioalwan: Addressed in fa385be81.

`MSIDAuthenticationScheme` now exposes `configureTelemetryEvent:`, which is a no-op by default. `MSIDAuthenticationSchemePop` overrides it to set `pop_key_source`, and `MSIDBaseRequestController` just calls:

```objc
[self.requestParameters.authScheme configureTelemetryEvent:event];
```

That removes the dynamic cast. Added unit coverage for base no-op plus internal/external PoP key-source values.

antrix1989 (approved):

PR #1937, opened 2026-08-12
[patch][bugfix] Avoid bound token provider symbol collision

antonioalwan:
## Summary
- Give the bound-token provider log prefix internal linkage.
- Prevent duplicate-symbol linker failures when MSAL and OneAuth are linked into the same app.
- Document the fix in `changelog.txt`.

## Validation
- `./build.py --targets mac_library --no-clean --no-xcpretty`

Split from #1934 so the linker fix can be reviewed and shipped independently.

### DIFF
--- a/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
+++ b/IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
@@ -28,7 +28,7 @@
 #import "MSIDLogger+Internal.h"
 #import "NSString+MSIDExtensions.h"
 
-NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
+static NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
 
 @implementation MSIDBoundTokenProvider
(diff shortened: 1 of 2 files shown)


sidudiki (approved):

sidudiki (approved):

IdentityCore/src/oauth2/token/MSIDBoundTokenProvider.m
```diff
@@ -28,7 +28,7 @@
 #import "MSIDLogger+Internal.h"
 #import "NSString+MSIDExtensions.h"
 
-NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
+static NSString *const MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX = @"[MSIDBoundTokenProvider]";
```
jasoncoolmax, line 31: Change looks right, and `MSIDLocalSPATokenAcquirer.m` already declares its prefix `static`, so this makes the two consistent. One thing: #1933 rewrites this same file and still has it non-static, so whoever merges second please make sure this doesn't get lost :)

antonioalwan:
No additional high-confidence correctness, security, concurrency, API-contract, or test-coverage issues were found.

PR #1938, opened 2026-08-14
[minor] [feature]: Validate broker nonce on the sourceApplication path behind a flight

hieunguyenmsft:
## Summary

`-[MSIDBrokerResponseHandler checkBrokerNonce:]` only compared `broker_nonce` when `sourceApplication` was `nil`. On the legacy `application:openURL:sourceApplication:` / `application:open:options:` path it returned `YES` without evaluating the nonce at all.

The original reasoning is sound as far as it goes: a non-nil `sourceApplication` is asserted by the OS, cannot be forged by the calling app, and `MSIDBrokerInteractiveController` already rejects any non-blank `sourceApplication` that isn't the known broker bundle id before this handler is reached.

But that establishes the *origin* of a response, not its *freshness*. Those address different things and aren't substitutes for one another. The nonce is the only freshness binding in the response envelope, and on this path it wasn't being checked.

## What changed

`checkBrokerNonce:` now compares the nonce first on every path:

- **Nonce matches** — accepted, unchanged.
- **Mismatch, `sourceApplication == nil`** — rejected, unchanged.
- **Mismatch, `sourceApplication != nil`** — gated by the new `enforce_broker_nonce` flight.
  - Flight **off (default)**: logs a warning and accepts. Identical to today's behavior.
  - Flight **on**: rejects with `MSIDErrorBrokerMismatchedResumeState`.

Defaulting off is deliberate. Any broker build that doesn't echo `broker_nonce` back would start failing sign-in the moment this is enforced, so the warning log lets us size that population per ring before flipping it on.

## Why this needs no broker-side change

The broker already writes `broker_nonce` into the response payload before encrypting it, so it travels inside the ciphertext and under the response hash. The client already generates it and hard-fails if it's missing from resume state. Both ends are in place — only the client-side comparison was conditional. No envelope format change, no protocol version bump, no coordinated broker release.

### `MSIDDefaultBrokerResponseHandlerTests` (5 new)

| `sourceApplication` | nonce | flight | expected |
|---|---|---|---|
| non-nil | mismatch | on | rejected |
| non-nil | missing | on | rejected |
| non-nil | mismatch | off | accepted (pins current behavior) |
| non-nil | matches | on | accepted (no happy-path regression) |
| nil | mismatch | off | rejected (flight doesn't weaken this path) |

### `MSIDLegacyBrokerResponseHandlerTests` (9 new)

`MSIDLegacyBrokerResponseHandler` calls `checkBrokerNonce:` twice — once on the decrypted success response, and once on the `error_code` branch — so the flight changes the ADAL path too. Both call sites are now covered.

Success path (decrypted response):

| `sourceApplication` | nonce | flight | expected |
|---|---|---|---|
| non-nil | mismatch | on | rejected |
| non-nil | missing | on | rejected |
| non-nil | mismatch | off | accepted (pins current behavior) |
| non-nil | matches | on | accepted (no happy-path regression) |
| nil | mismatch | off | rejected (flight doesn't weaken this path) |

`error_code` path:

| `sourceApplication` | nonce | flight | expected |
|---|---|---|---|
| non-nil | mismatch | on | rejected with nonce error |
| non-nil | missing | on | rejected with nonce error |
| non-nil | mismatch | off | underlying broker error surfaces (pins current behavior) |
| non-nil | matches | on | underlying broker error surfaces (no regression) |

The "flight off" rows matter most here: they prove enabling the flight is the only thing that can change ADAL behavior, and that with it off the ADAL path is bit-for-bit what it is today.

`tearDown` in both suites now resets `MSIDFlightManager.sharedInstance.flightProvider`, since it's a singleton and would otherwise leak across tests.

Because the flight defaults off, the existing tests that encode today's permissive behavior are untouched and still pass.

**Results:** 40 pass in `MSIDDefaultBrokerResponseHandlerTests` (5 new, 35 existing), 28 pass in `MSIDLegacyBrokerResponseHandlerTests` (9 new, 19 existing), 8 pass in `MSIDBrokerResponseHandlerTests`. No regressions. `MSIDFlightManagerMockProvider` already builds into `IdentityTest iOS`, which the test target links, so no `.pbxproj` change was needed.

## Note for reviewers

`ADBrokerAADv1Response` returns failure responses unencrypted (`shouldEncryptFailureResponse == NO`), which is exactly the `error_code` branch in `MSIDLegacyBrokerResponseHandler`. The nonce there is read from cleartext query params and is editable in transit, so enforcement on that branch stops a blind replay where the attacker only holds a captured response, but not an attacker who can observe the outbound request. It's covered by tests for behavioral completeness, not because it carries a real guarantee. Closing that gap properly needs an envelope change and is out of scope here.

### DIFF
--- a/IdentityCore/src/MSIDConstants.h
+++ b/IdentityCore/src/MSIDConstants.h
@@ -301,6 +301,13 @@ extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_OPEN_NEW_WINDOW_IN_BROWSER;
 /// Default: OFF
 extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_MOBILE_ONBOARDING;
 
+/// Enforces broker_nonce validation on broker responses even when sourceApplication is available.
+/// sourceApplication establishes the origin of a response, not its freshness, so it is not a
+/// substitute for the nonce. Flighted so impact on brokers that do not echo the nonce back can be
+/// measured before enforcement is turned on.
+/// Default: OFF
+extern NSString * _Nonnull const MSID_FLIGHT_ENFORCE_BROKER_NONCE;
+
 /// Flight key for MDM profile install notification delay (seconds).
 /// Owner: swagup
 extern NSString * _Nonnull const MSID_FLIGHT_MDM_PROFILE_INSTALLED_NOTIFICATION_DELAY;
--- a/IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
+++ b/IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
@@ -308,7 +308,9 @@
         return NO;
     }
 
-    NSString *nonceMissing = responseDict[@"broker_nonce"] ? @"NO" : @"YES";
+    // A response carrying "broker_nonce=" decodes to a present-but-empty string, so presence alone
+    // would report such a broker as having echoed a nonce. Treat nil and blank alike.
+    NSString *nonceMissing = [NSString msidIsStringNilOrBlank:responseDict[@"broker_nonce"]] ? @"YES" : @"NO";
 
     if (![[MSIDFlightManager sharedInstance] boolForKey:MSID_FLIGHT_ENFORCE_BROKER_NONCE])
     {
--- a/IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
+++ b/IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
@@ -41,6 +41,7 @@
 #import "MSIDAuthenticationSchemePop.h"
 #import "MSIDAuthenticationScheme.h"
 #import "MSIDAuthScheme.h"
+#import "MSIDFlightManager.h"
 
 @interface MSIDBrokerResponseHandler()
 
@@ -291,13 +292,32 @@
 
 - (BOOL)checkBrokerNonce:(NSDictionary *)responseDict
 {
-    // only verify nonce if sourceApplication is nil
+    if ([self.brokerNonce isEqualToString:responseDict[@"broker_nonce"]])
+    {
+        return YES;
+    }
+
+    // Historically the nonce was verified only when sourceApplication was nil, on the basis that a
+    // non-nil sourceApplication is asserted by the OS and is validated against the broker bundle id
+    // before this handler is reached. sourceApplication establishes the origin of a response, not its
+    // freshness, so it does not protect against replay of a previously captured response. Enforcement
+    // on that path is flighted so that impact on brokers which do not echo the nonce back can be
+    // measured before it is turned on.
     if (!self.sourceApplicationAvailable)
     {
-        return [self.brokerNonce isEqualToString:responseDict[@"broker_nonce"]];
+        return NO;
     }
-    
-    return YES;
+
+    NSString *nonceMissing = responseDict[@"broker_nonce"] ? @"NO" : @"YES";
+
+    if (![[MSIDFlightManager sharedInstance] boolForKey:MSID_FLIGHT_ENFORCE_BROKER_NONCE])
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, nil, @"Broker nonce mismatch on the sourceApplication code path, allowed because enforcement is disabled. Nonce missing from response: %@", nonceMissing);
+        return YES;
+    }
+
+    MSID_LOG_WITH_CTX(MSIDLogLevelError, nil, @"Broker nonce mismatch on the sourceApplication code path, rejecting response. Nonce missing from response: %@", nonceMissing);
+    return NO;
 }
 
 #pragma mark - Abstract
--- a/IdentityCore/src/MSIDConstants.m
+++ b/IdentityCore/src/MSIDConstants.m
@@ -136,6 +136,8 @@ NSString *const MSID_FLIGHT_DISABLE_OPEN_NEW_WINDOW_IN_BROWSER = @"disable_open_
 
 NSString *const MSID_FLIGHT_DISABLE_MOBILE_ONBOARDING = @"disable_mobile_onboarding";
 
+NSString *const MSID_FLIGHT_ENFORCE_BROKER_NONCE = @"enforce_broker_nonce";
+
 NSString *const MSID_FLIGHT_MDM_PROFILE_INSTALLED_NOTIFICATION_DELAY = @"mdm_profile_installed_notification_delay";
 
 NSTimeInterval const MSIDMDMProfileInstalledNotificationDefaultDelay = 75.0;
--- a/IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
+++ b/IdentityCore/tests/integration/ios/MSIDDefaultBrokerResponseHandlerTests.m
@@ -48,6 +48,8 @@
 #import "MSIDDefaultTokenCacheAccessor.h"
 #import "MSIDAccountMetadataCacheAccessor.h"
 #import "MSIDOnboardingBlobFieldKeys.h"
+#import "MSIDFlightManager.h"
+#import "MSIDFlightManagerMockProvider.h"
 
 @interface MSIDDefaultBrokerResponseHandlerTests : XCTestCase
 
@@ -77,6 +79,8 @@
     SecItemDelete((CFDictionaryRef)query);
     
     [[NSUserDefaults standardUserDefaults] removeObjectForKey:MSID_BROKER_RESUME_DICTIONARY_KEY];
+    
+    MSIDFlightManager.sharedInstance.flightProvider = nil;
     
     [super tearDown];
 }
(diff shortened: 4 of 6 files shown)


IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
```diff
@@ -291,13 +292,32 @@ - (NSDictionary *)verifyResumeStateDictionary:(NSURL *)response error:(NSError *
...
     {
-        return [self.brokerNonce isEqualToString:responseDict[@"broker_nonce"]];
+        return NO;
     }
-    
-    return YES;
+
+    NSString *nonceMissing = responseDict[@"broker_nonce"] ? @"NO" : @"YES";
+
+    if (![[MSIDFlightManager sharedInstance] boolForKey:MSID_FLIGHT_ENFORCE_BROKER_NONCE])
```
jasoncoolmax, line 315: `MSIDLegacyBrokerResponseHandler` also calls `checkBrokerNonce` twice (once on the decrypted response, once on the `error_code` path with `encryptedParams`), so this flight changes the ADAL path too. Tests only cover `MSIDDefaultBrokerResponseHandler` here, shall we add coverage for the legacy handler as well?

hieunguyenmsft: Good catch — added in 44aaa28. Nine tests for `MSIDLegacyBrokerResponseHandlerTests`, covering both call sites.

**Success path** (decrypted response, line 95): mismatch/on → rejected, missing/on → rejected, mismatch/off → accepted, match/on → accepted, and `sourceApplication == nil` + mismatch/off → rejected.

**`error_code` path** (line 104): mismatch/on and missing/on → nonce error, mismatch/off and match/on → underlying broker error (code 213) surfaces unchanged.

The "flight off" rows are the ones that carry the weight for your concern: they pin ADAL behavior as bit-for-bit identical to today, so enabling the flight is the only thing that can change it.

One thing worth calling out while we're here: line 104 reads the nonce from `encryptedParams`, i.e. the raw query parameters. `ADBrokerAADv1Response` returns failure responses unencrypted (`shouldEncryptFailureResponse == NO`), so on that branch the nonce is in cleartext and editable in transit. Enforcement there stops a blind replay where the attacker only holds a captured response, but not one who can observe the outbound request. It's covered for behavioral completeness, not because it carries a real guarantee — I've noted that in the PR description so it isn't mistaken for coverage this change provides.

Full run: 28 pass in `MSIDLegacyBrokerResponseHandlerTests` (9 new, 19 existing), 40 in `MSIDDefaultBrokerResponseHandlerTests`, 8 in `MSIDBrokerResponseHandlerTests`. No regressions.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
```diff
@@ -291,13 +292,32 @@ - (NSDictionary *)verifyResumeStateDictionary:(NSURL *)response error:(NSError *
...
+    {
+        MSID_LOG_WITH_CTX(MSIDLogLevelWarning, nil, @"Broker nonce mismatch on the sourceApplication code path, allowed because enforcement is disabled. Nonce missing from response: %@", nonceMissing);
```
josephpab, line 317: The rollout plan is to watch this warning's volume and then enable per ring, but this string is identical across all three call sites: the decrypted success payload in `MSIDDefaultBrokerResponseHandler`, the decrypted success payload in `MSIDLegacyBrokerResponseHandler`, and the cleartext `error_code` branch in `MSIDLegacyBrokerResponseHandler`.

Those are different broker serialization paths and could easily have different populations, and they carry very different risk when the flight flips on. As written the logs can't tell them apart, so the data we collect won't answer the question we're collecting it for.

Can we pass a discriminator into `checkBrokerNonce:` (or take the caller's context/correlation id) and include it here? Something like `checkBrokerNonce:responseDict source:@"legacy_error_code"` would make the warning counts actually segmentable.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
```diff
@@ -291,13 +292,32 @@ - (NSDictionary *)verifyResumeStateDictionary:(NSURL *)response error:(NSError *
...
+    // measured before it is turned on.
     if (!self.sourceApplicationAvailable)
     {
-        return [self.brokerNonce isEqualToString:responseDict[@"broker_nonce"]];
+        return NO;
     }
-    
-    return YES;
+
+    NSString *nonceMissing = responseDict[@"broker_nonce"] ? @"NO" : @"YES";
```
josephpab: This misclassifies an empty nonce. `msidDictionaryFromWWWFormURLEncodedString` defaults `value` to `@""` when a key has no value (`NSDictionary+MSIDExtensions.m`), so a broker sending `broker_nonce=` lands here as a present-but-empty string and gets reported as `nonceMissing = NO`.

That's exactly backwards for the signal this flight exists to collect: a broker that isn't really echoing a nonce would show up in the "sent one, it mismatched" bucket instead of the "didn't send one" bucket.

`[NSString msidIsStringNilOrBlank:responseDict[@"broker_nonce"]] ? @"YES" : @"NO"` covers both.

IdentityCore/src/MSIDConstants.h
```diff
@@ -301,6 +301,13 @@ extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_OPEN_NEW_WINDOW_IN_BROWSER;
 /// Default: OFF
 extern NSString * _Nonnull const MSID_FLIGHT_DISABLE_MOBILE_ONBOARDING;
 
+/// Enforces broker_nonce validation on broker responses even when sourceApplication is available.
+/// sourceApplication establishes the origin of a response, not its freshness, so it is not a
+/// substitute for the nonce. Flighted so impact on brokers that do not echo the nonce back can be
+/// measured before enforcement is turned on.
+/// Default: OFF
+extern NSString * _Nonnull const MSID_FLIGHT_ENFORCE_BROKER_NONCE;
```
josephpab, line 309: One flight gates two changes that I don't think should ship together.

On the success path a mismatch turns a working sign-in into a failure — bad, but the failure is honest. On the `error_code` branch in `MSIDLegacyBrokerResponseHandler` it does something different: it replaces a real broker error with a generic one. Your own test `testHandleBrokerResponse_whenBrokerErrorResponse_andSourceApplicationNonNil_andNonceMismatch_andEnforcementFlightOff_shouldReturnUnderlyingBrokerError` pins that today ADAL gets `ADAuthenticationErrorDomain` / 213 / `consent_required`; with the flight on it gets `MSIDErrorBrokerMismatchedResumeState` instead. Apps branch on those codes to decide whether to re-prompt, so we'd be converting a recoverable interactive case into a dead end.

And the two aren't the same population. The encrypted success payload and the unencrypted error response are built separately in the broker, so a build could echo the nonce in one and not the other. With a single flight there's no way to validate one before the other — and per my other comment the logs won't distinguish them either.

Given your own note that enforcement on that branch doesn't buy a real guarantee anyway, could we either add a second flight for the `error_code` branch, or just leave that call site unenforced in this PR?

Nit while you're here: most of the recent flights in this file carry `/// Owner:`. Worth adding.

josephpab (commented):
Nice writeup, and I agree with the core premise: `sourceApplication` proves origin, not freshness, and the nonce was the only freshness binding we were skipping. The flight-off-by-default shape is right, and the tests that pin today's permissive behavior are the right way to prove this is inert until we flip it.

Three things before I'd want to enable it anywhere. The main one is that the `error_code` branch and the success path are bundled under one flight — see the comment on `MSIDConstants.h`. The other two are about the rollout telemetry, which is the whole justification for shipping this off, so I'd like it to be accurate before we start reading it.

Nothing here is a correctness problem with the flight off; the diff is behavior-preserving in the default configuration and I verified that path by hand.

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
hieunguyenmsft: just one flying plan for feature control, I updated description

IdentityCore/src/requests/sdk/MSIDBrokerResponseHandler.m
hieunguyenmsft: Good catch — confirmed and fixed in fbb727e76.

`msidDictionaryFromWWWFormURLEncodedString` initializes `NSString *value = @""` (`NSDictionary+MSIDExtensions.m:70`) and only overwrites it when `queryElements.count == 2`, so both `broker_nonce=` and a bare `broker_nonce` land in the dictionary as a non-nil empty string. The old presence check reported those as `nonceMissing = NO`, which is precisely backwards for the bucket this flight is meant to measure.

Now:

```objc
NSString *nonceMissing = [NSString msidIsStringNilOrBlank:responseDict[@"broker_nonce"]] ? @"YES" : @"NO";
```

`msidIsStringNilOrBlank` also covers the whitespace-only case (the decoder trims before decoding, so `broker_nonce=%20` arrives as a blank string), which plain `length` would have missed.

No behavior change — `nonceMissing` only feeds the two log statements, and control flow is unchanged, so I didn't add a test for it rather than assert on log text. Existing coverage still green: 68 tests across `MSIDDefaultBrokerResponseHandlerTests` and `MSIDLegacyBrokerResponseHandlerTests`, 0 failures.

josephpab (dismissed):

josephpab (approved):

PR #1943, opened 2026-08-18
[patch] [bugfix]: Rename execution flow tag helpers to avoid duplicate symbols with OneAuth

josephpab:
Same change as #1942, targeted at `dev` so the fix does not regress after the release branch merges.

IdentityCore exports 8 symbols that OneAuth 11.1.1's vendored IdentityCore copy also exports **unprefixed**. Linking `libMSAL.a` and `libOneAuth.a` into the same binary fails:

```
ld: 8 duplicate symbols
clang: error: linker command failed with exit code 1
```

### Cause

OneAuth vendors a copy of IdentityCore and renames types/classes/selectors to `MSAI*`, but leaves free functions and file-scope globals unmangled — e.g. in their copy:

```objc
NSString *MSIDExecutionFlowNetworkTagToString(MSAIMSIDExecutionFlowNetworkTag state)
//        ^^^^ not renamed                    ^^^^ renamed
```

so those names stay identical to ours and collide at link time.

### Change

Rename the 7 colliding helpers to `MSIDStringFrom<Tag>()`, keeping the `MSID` prefix so OneAuth's rename tooling picks them up going forward:

| Old | New |
|---|---|
| `MSIDExecutionFlowNetworkTagToString` | `MSIDStringFromExecutionFlowNetworkTag` |
| `MSIDTokenRequestTagToString` | `MSIDStringFromTokenRequestTag` |
| `MSIDRequestControllerFactoryTagToString` | `MSIDStringFromRequestControllerFactoryTag` |
| `MSIDSSORemoteInteractiveTokenRequestTagToString` | `MSIDStringFromSSORemoteInteractiveTokenRequestTag` |
| `MSIDSSORemoteSilentTokenRequestTagToString` | `MSIDStringFromSSORemoteSilentTokenRequestTag` |
| `MSIDCloudInstanceHostNameTagToString` | `MSIDStringFromCloudInstanceHostNameTag` |
| `MSIDPkeyAuthTagToString` | `MSIDStringFromPkeyAuthTag` |

The 8th symbol, `MSID_BOUND_TOKEN_PROVIDER_LOG_PREFIX`, is made `static`. It is used only within `MSIDBoundTokenProvider.m` and has no header declaration, so it never needed external linkage.

`MSIDCustomHeaderTagToString` is intentionally **not** renamed — it landed after OneAuth's vendored snapshot and does not collide.

### Notes

- Pure rename: 16 files, 79 insertions / 79 deletions, no behavior change.
- A fix for OneAuth's rename script exists on their `dev` but is not yet in `master`; this change is compatible either way.

### DIFF
--- a/IdentityCore/src/controllers/MSIDRequestControllerFactory.m
+++ b/IdentityCore/src/controllers/MSIDRequestControllerFactory.m
@@ -50,7 +50,7 @@
                                                 tokenRequestProvider:(id<MSIDTokenRequestProviding>)tokenRequestProvider
                                                                error:(NSError *__autoreleasing*)error
 {
-    MSIDExecutionFlowInsertTag(MSIDRequestControllerFactoryTagToString(MSIDSilentControllerForParametersTag),
+    MSIDExecutionFlowInsertTag(MSIDStringFromRequestControllerFactoryTag(MSIDSilentControllerForParametersTag),
                                    @{MSID_EXECUTION_FLOW_DIAGNOSTIC_ID:@(parameters.xpcMode)},
                                    parameters.correlationId);
     if (parameters.xpcMode == MSIDXpcModeDisabled)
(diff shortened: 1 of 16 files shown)


josephpab:
Superseded — consolidating onto a single branch/commit (`josephpab/fix-duplicate-symbols-oneauth`, b3cbf0d5a) so `dev` and `release/2.15.0` get the identical commit rather than two divergent ones. Re-opening the dev PR from that branch.

PR #1946, opened 2026-08-20
[patch][bugfix]: Prevent extraQueryParameters from overriding reserved OAuth2 parameters

mipetriu:
Reserved keys are now stripped from both `extraAuthorizeURLQueryParameters` and `extraURLQueryParameters` before the merge, so SDK-owned values always win. Establishes the invariant: **EQP can never override a parameter the SDK itself sets.**

Reserved set: `client_id`, `redirect_uri`, `response_type`, `scope`, `state`, `code_challenge`, `code_challenge_method`, `post_logout_redirect_uri`.

Also adds the `reject_reserved_extra_query_parameters` flight (**default OFF**). When enabled, requests carrying reserved keys fail validation up front with `MSIDErrorInvalidDeveloperParameter` instead of being silently filtered — matching MSAL.NET's `DuplicateQueryParameterError` behavior.

| Change | File |
|---|---|
| Reserved-key set + case-insensitive filter helpers | `MSIDRequestParameters.h/.m` |
| Strip reserved keys before merging EQP (authorize) | `MSIDInteractiveTokenRequestParameters.m` |
| Strip reserved keys before merging EQP (logout) | `MSIDWebviewFactory.m` |
| Flight-gated hard `NSError` validation | `MSIDInteractiveTokenRequestParameters.m` |
| `reject_reserved_extra_query_parameters` flight constant | `MSIDConstants.h/.m` |

## Design notes

**`nonce` is deliberately NOT reserved**, despite the bug asking for it "at minimum." IdentityCore never writes `nonce` into the authorize URL — the only match in the repo is an unrelated error string in `MSIDAADWebviewFactory.m`. Blocking it would break apps doing OIDC hybrid flows while protecting nothing.

**Token-endpoint query params are not filtered.** That's the `/token` URL query, not `/authorize`; the server reads the POST body. Verified MSAL.NET does the same — its `CheckForDuplicateQueryParameters` is called from exactly one place (the authorize URL build in `AuthCodeRequestComponent.cs`), while `TokenClient.SendHttpAndClearTelemetryAsync` appends EQP to the token endpoint with no check at all, because reserved token params travel in the POST body via `AddBodyParameter`.

**Matching is case-insensitive.** ESTS may treat query-param names case-insensitively, so a key like `Client_Id` would not collide in the `NSDictionary` and would emit a *duplicate* parameter into the URL. Lowercasing on compare closes that bypass. (Worth noting MSAL.NET has this gap — it uses `new Dictionary<string, string>` with no comparer, so its check is ordinal case-sensitive.)

**`allowAnyExtraURLQueryParameters` is honored as an escape hatch** by both the filter and the validation. It is set to `YES` by `MSIDInteractiveTokenRequestParameters+BrowserNativeMessageGetToken.m` for browser-native / nested-auth (double-broker); without the exemption that flow breaks.

**Filtering rather than reordering the writes.** `MSIDAADV1WebviewFactory.m` deliberately re-sets `result[MSID_OAUTH2_SCOPE]` after calling `super`, so a naive "just write SDK params last" fix would be unsafe across the subclass hierarchy.

Only key names are logged on the strip path — values are caller-controlled and may carry PII.

## Compatibility

Non-breaking by default. Apps passing reserved keys today silently lose them (previously they silently won) and get a warning log. The hard failure is opt-in via the flight.

## Testing

Full iOS suite: **3118 tests, 0 failures** (11 new).

New coverage in `MSIDWebviewFactoryTests.m`:
- Reserved keys in `extraAuthorizeURLQueryParameters` → SDK values retained
- Reserved keys in `extraURLQueryParameters` → SDK values retained (incl. PKCE + state)
- Case variant `Client_Id` → stripped
- `allowAnyExtraURLQueryParameters = YES` → reserved keys pass through
- Logout path: `state` + `post_logout_redirect_uri` → SDK values retained
- Non-reserved params (`eqp1`, `domain_hint`) → still forwarded

New coverage in `MSIDInteractiveTokenRequestParametersTests.m`:
- Flight OFF + reserved key → validation succeeds
- Flight ON + reserved keys → fails with `MSIDErrorInvalidDeveloperParameter`, message names the offending keys
- Flight ON + `allowAnyExtraURLQueryParameters` → succeeds
- Flight ON + only non-reserved keys → succeeds
- `tearDown` resets `MSIDFlightManager.sharedInstance.flightProvider`


IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -374,6 +375,81 @@ - (BOOL)validateParametersWithError:(NSError *__autoreleasing*)error
...
+                                      MSID_OAUTH2_SIGNOUT_REDIRECT_URI];
+
+        NSMutableSet<NSString *> *lowercasedKeys = [NSMutableSet setWithCapacity:keys.count];
+
+        for (NSString *key in keys)
+        {
+            [lowercasedKeys addObject:key.lowercaseString];
+        }
+
+        reservedKeys = lowercasedKeys;
```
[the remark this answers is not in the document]
mipetriu: Good catch — fixed in e3ea42ae4.

`reservedURLQueryParameterKeys` now stores `[lowercasedKeys copy]` in the static. This was the one that actually mattered: the set is shared process-wide, so a cast-and-mutate would have changed reserved-key behavior for every request in the process.

Also tightened the two per-call returns you flagged as lower severity — `reservedKeysInParameters:` (line 428) and `parametersByRemovingReservedKeys:`, which was returning its `mutableCopy` directly through an `NSDictionary` declaration. Same contract issue, so it made sense to clean up together.

No behavior change: all four call sites treat the returns as read-only (`addEntriesFromDictionary:` and `unionSet:`, where the receiver is a separate mutable object). `MSIDWebviewFactoryTests` and `MSIDInteractiveTokenRequestParametersTests` still pass.

antonioalwan:

antonioalwan (approved):

IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -374,6 +375,81 @@ - (BOOL)validateParametersWithError:(NSError *__autoreleasing*)error
...
+    }
+
+    NSSet<NSString *> *reservedKeys = [self reservedURLQueryParameterKeys];
+    NSMutableSet<NSString *> *matchedKeys = [NSMutableSet new];
+
+    for (id key in parameters)
+    {
+        if (![key isKindOfClass:[NSString class]]) continue;
+
+        if ([reservedKeys containsObject:[(NSString *)key lowercaseString]])
```
antrix1989, line 422: Can we normalize the key the same way as the URL serializer before checking this set? `@" client_id "` survives this lowercase-only comparison, including flight-on validation, but serialization trims it and emits a second `client_id` alongside the SDK value. The existing `msidNormalizedString` provides trimming and lowercasing; keep the original key in `matchedKeys` so removal still works. Please test padded keys against the final URL's query items, preserving duplicates, and verify flight-on rejection as well.

IdentityCore/src/oauth2/MSIDWebviewFactory.m
```diff
@@ -230,7 +230,15 @@ @implementation MSIDWebviewFactory
     result[MSID_OAUTH2_STATE] = state.msidBase64UrlEncode;
     [result addEntriesFromDictionary:[self metadataFromRequestParameters:parameters]];
     [result addEntriesFromDictionary:parameters.appRequestMetadata];
-    [result addEntriesFromDictionary:parameters.extraURLQueryParameters];
+
+    // Reserved OAuth2 parameters are dropped so that caller supplied extra query parameters cannot
+    // override the sign out redirect URI or the request state validated on the response.
+    NSDictionary *extraURLQueryParameters = [parameters parametersByRemovingReservedKeys:parameters.extraURLQueryParameters];
```
antrix1989, line 236: Can we apply the reserved-key rejection check on the signout path before network or browser work? The new flight contract includes logout, but its check exists only in `MSIDInteractiveTokenRequestParameters`. Signout uses `MSIDInteractiveRequestParameters` and reaches this filter without that validation, so flight-on requests silently continue instead of failing upfront. Please share only the reserved-EQP validation, not token-specific requirements, and add real signout-parameter tests for flight on/off and the allow-any exemption.

IdentityCore/src/parameters/MSIDRequestParameters.m
```diff
@@ -374,6 +375,81 @@ - (BOOL)validateParametersWithError:(NSError *__autoreleasing*)error
...
+
++ (NSSet<NSString *> *)reservedURLQueryParameterKeys
+{
+    static NSSet<NSString *> *reservedKeys;
+    static dispatch_once_t onceToken;
+
+    dispatch_once(&onceToken, ^{
+        NSArray<NSString *> *keys = @[MSID_OAUTH2_CLIENT_ID,
+                                      MSID_OAUTH2_REDIRECT_URI,
+                                      MSID_OAUTH2_RESPONSE_TYPE,
```
antrix1989, line 388: This also strips the broker-owned `response_type=none` used by MSA QR account transfer. That flow does not set `allowAnyExtraURLQueryParameters`, so this merge keeps `response_type=code` instead, even with the flight off. Can we carry the trusted QR response type separately while still blocking arbitrary developer overrides? Please add a regression for the QR request and an ordinary override.

IdentityCore/tests/MSIDWebviewFactoryTests.m
```diff
@@ -164,6 +164,105 @@ - (void)testLogoutParametersFromRequestParameters_whenValidParameters_shouldRetu
...
+        MSID_OAUTH2_CLIENT_ID : @"attacker_client_id",
+        MSID_OAUTH2_STATE : @"attacker_state",
+        MSID_OAUTH2_CODE_CHALLENGE : @"attacker_challenge",
+        MSID_OAUTH2_CODE_CHALLENGE_METHOD : @"plain",
+        @"domain_hint" : @"contoso.com"
+    };
+
+    NSString *requestState = @"state";
+    MSIDPkce *pkce = [MSIDPkce new];
+
```
antrix1989, line 210: Could we also test this with `pkce:nil`? When PKCE is present, the factory already overwrites both challenge fields after merging extras, so the PKCE assertions still pass if these two keys are no longer filtered. Please cover each extra-parameter dictionary with PKCE disabled and assert that `code_challenge` and `code_challenge_method` are absent while nonreserved extras are preserved.

PR #1949, opened 2026-08-25
[patch] [bugfix]: Do not send external_key_pop to the /token endpoint

antonioalwan:
## Problem

`MSIDTokenRequest` copies the **entire** `authScheme.schemeParameters` dictionary into the `/token` request body:

```objc
NSDictionary *authHeaders = authScheme.schemeParameters;
if ([authHeaders count] > 0)
{
    [parameters addEntriesFromDictionary:authHeaders];
}
```

For external-key AT PoP (#1934), `MSALAuthenticationSchemePop.getSchemeParameters:` adds `external_key_pop=1` to that dictionary. So on the direct (non-broker) path, eSTS receives `token_type`, `req_cnf` **and** `external_key_pop`, even though `external_key_pop` is a client-internal marker that only the broker consumes (`MSIDBrokerTokenRequest` query dict and resume dict).

The broker already strips it before calling eSTS, in `ADBrokerAADV2TokenRequest`:

```objc
NSMutableDictionary *authHeaders = [authScheme.schemeParameters mutableCopy];
[authHeaders removeObjectForKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
```

So the two paths disagreed on what goes on the wire.

## Fix

Rather than duplicating the removal at each request site, the scheme now owns the distinction between its full parameter set and the subset that is a wire parameter.

- `MSIDAuthenticationScheme` gains `tokenEndpointParameters`, defaulting to `schemeParameters`.
- `MSIDAuthenticationSchemePop` overrides it to drop `external_key_pop`.
- `MSIDTokenRequest` uses `tokenEndpointParameters`, which covers `MSIDAuthorizationCodeGrantRequest`, `MSIDRefreshTokenGrantRequest`, and `MSIDDeviceTokenGrantRequest` through inheritance.

`schemeParameters` is deliberately unchanged, so broker request/resume state, JSON serialization, `isExternalKeyPop`, and telemetry all keep seeing the marker.

## Tests

Added to `MSIDAuthenticationSchemePopTest`:

- `testTokenEndpointParameters_whenExternalKeyPop_shouldExcludeExternalKeyMarker`
- `testTokenEndpointParameters_whenInternalKeyPop_shouldReturnSchemeParameters`
- `testTokenRequestParameters_whenExternalKeyPop_shouldNotSendExternalKeyMarkerToTokenEndpoint`

`IdentityCoreTests Mac/MSIDAuthenticationSchemePopTest`: 16 tests, 0 failures.

## Notes

Bearer and SSH-cert schemes are unaffected, they inherit the default and return `schemeParameters` unchanged.

### DIFF
--- a/IdentityCore/src/auth_scheme/MSIDAuthenticationScheme.m
+++ b/IdentityCore/src/auth_scheme/MSIDAuthenticationScheme.m
@@ -60,6 +60,11 @@
     return MSIDAuthSchemeBearer;
 }
 
+- (NSDictionary *)tokenEndpointParameters
+{
+    return self.schemeParameters;
+}
+
 - (MSIDCredentialType)credentialType
 {
     return MSIDAccessTokenType;
--- a/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
+++ b/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
@@ -150,7 +150,7 @@
     // external_key_pop is a client-internal marker consumed by the broker, not a '/token' parameter.
     NSMutableDictionary *parameters = [self.schemeParameters mutableCopy];
     [parameters removeObjectForKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
-    return parameters;
+    return [parameters copy];
 }
 
 #if !EXCLUDE_FROM_MSALCPP
--- a/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
+++ b/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
@@ -140,6 +140,19 @@
     return [self.externalKeyPop isEqualToString:@"1"];
 }
 
+- (NSDictionary *)tokenEndpointParameters
+{
+    if (!self.externalKeyPop)
+    {
+        return self.schemeParameters;
+    }
+
+    // external_key_pop is a client-internal marker consumed by the broker, not a '/token' parameter.
+    NSMutableDictionary *parameters = [self.schemeParameters mutableCopy];
+    [parameters removeObjectForKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
+    return parameters;
+}
+
 #if !EXCLUDE_FROM_MSALCPP
 - (void)configureTelemetryEvent:(MSIDTelemetryAPIEvent *)event
 {
--- a/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
+++ b/IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
@@ -142,11 +142,6 @@
 
 - (NSDictionary *)tokenEndpointParameters
 {
-    if (!self.externalKeyPop)
-    {
-        return self.schemeParameters;
-    }
-
     // external_key_pop is a client-internal marker consumed by the broker, not a '/token' parameter.
     NSMutableDictionary *parameters = [self.schemeParameters mutableCopy];
     [parameters removeObjectForKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
--- a/IdentityCore/src/network/request/MSIDTokenRequest.m
+++ b/IdentityCore/src/network/request/MSIDTokenRequest.m
@@ -48,7 +48,7 @@
         parameters[MSID_OAUTH2_CLIENT_ID] = clientId;
         parameters[MSID_OAUTH2_SCOPE] = scope;
         
-        NSDictionary *authHeaders = authScheme.schemeParameters;
+        NSDictionary *authHeaders = authScheme.tokenEndpointParameters;
         if ([authHeaders count] > 0)
         {
             [parameters addEntriesFromDictionary:authHeaders];
--- a/IdentityCore/src/network/request/MSIDTokenRequest.m
+++ b/IdentityCore/src/network/request/MSIDTokenRequest.m
@@ -48,10 +48,10 @@
         parameters[MSID_OAUTH2_CLIENT_ID] = clientId;
         parameters[MSID_OAUTH2_SCOPE] = scope;
         
-        NSDictionary *authHeaders = authScheme.tokenEndpointParameters;
-        if ([authHeaders count] > 0)
+        NSDictionary *schemeBodyParameters = authScheme.tokenEndpointParameters;
+        if ([schemeBodyParameters count] > 0)
         {
-            [parameters addEntriesFromDictionary:authHeaders];
+            [parameters addEntriesFromDictionary:schemeBodyParameters];
         }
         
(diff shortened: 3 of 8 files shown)


IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
```diff
@@ -140,6 +140,19 @@ - (BOOL)isExternalKeyPop
...
+{
+    if (!self.externalKeyPop)
+    {
+        return self.schemeParameters;
+    }
+
+    // external_key_pop is a client-internal marker consumed by the broker, not a '/token' parameter.
+    NSMutableDictionary *parameters = [self.schemeParameters mutableCopy];
+    [parameters removeObjectForKey:MSID_OAUTH2_EXTERNAL_KEY_POP];
+    return parameters;
```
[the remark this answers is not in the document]
antonioalwan: Good catch, fixed in 3924eee. The override now returns `[parameters copy]`, so callers get an immutable `NSDictionary` and cannot mutate it by casting. The internal-key path already returned the immutable `schemeParameters` dictionary, so both paths are now consistent.

IdentityCore/tests/MSIDAuthenticationSchemePopTest.m
```diff
@@ -27,6 +27,8 @@
 #import "MSIDAccessTokenWithAuthScheme.h"
 #import "MSIDTelemetryAPIEvent.h"
 #import "MSIDTelemetryEventStrings.h"
+#import "MSIDOAuth2Constants.h"
+#import "MSIDTokenRequest.h"
```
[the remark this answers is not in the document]
antonioalwan: Not applicable here, this is a test file. `EXCLUDE_FROM_MSALCPP` is not defined anywhere in this repo (no `.pbxproj` or `.xcconfig` sets it); it is set by the MSAL C++ consumer, which compiles `IdentityCore/src` only and never `IdentityCore/tests`. The established convention in this directory matches that: `MSIDDeviceTokenGrantRequestTests.m`, `MSIDDeviceTokenGrantRequestNetworkTests.m`, `MSIDAADV2Oauth2FactoryTests.m`, and `mocks/MSIDDeviceTokenGrantRequestMock.h` all import `MSIDTokenRequest` subclasses unconditionally. Guarding this one test would be inconsistent with the other four.

IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
```diff
@@ -140,6 +140,19 @@ - (BOOL)isExternalKeyPop
     return [self.externalKeyPop isEqualToString:@"1"];
 }
 
+- (NSDictionary *)tokenEndpointParameters
+{
+    if (!self.externalKeyPop)
```
josephpab: This branches on `self.externalKeyPop`, but the thing that actually needs stripping is the key inside `self.schemeParameters`. Those two are only guaranteed to agree because both are populated in `initWithSchemeParameters:` (and both carried over in `copyWithZone:`). It's correct today, but it's an invariant someone has to keep holding.

Dropping the early return and always doing `mutableCopy` + `removeObjectForKey:` is identical in behavior -- `removeObjectForKey:` is a no-op when the key is absent -- and removes the coupling. The cost is one dictionary copy on the internal-PoP path, which happens once per token request, so probably not worth optimizing for.

Not blocking either way, just flagging that the guard is buying very little here.

IdentityCore/src/network/request/MSIDTokenRequest.m
```diff
@@ -48,7 +48,7 @@ - (instancetype)initWithEndpoint:(NSURL *)endpoint
         parameters[MSID_OAUTH2_CLIENT_ID] = clientId;
         parameters[MSID_OAUTH2_SCOPE] = scope;
         
-        NSDictionary *authHeaders = authScheme.schemeParameters;
+        NSDictionary *authHeaders = authScheme.tokenEndpointParameters;
```
josephpab: `authHeaders` was already slightly off, but now that the right-hand side literally says `tokenEndpointParameters` the mismatch is more obvious -- these land in `_parameters` and go out as the form-encoded POST body, not as headers.

Since the line is being touched anyway, worth renaming to `schemeParameters` or `schemeBodyParameters`.

IdentityCore/src/auth_scheme/MSIDAuthenticationScheme.m
```diff
@@ -60,6 +60,11 @@ - (MSIDAuthScheme)authSchemeFromParameters:(__unused NSDictionary *)schemeParame
     return MSIDAuthSchemeBearer;
 }
 
+- (NSDictionary *)tokenEndpointParameters
```
josephpab, line 63: The description says Bearer and SSH-cert are unaffected because they inherit this default, but nothing asserts it -- every new test in this PR is on `MSIDAuthenticationSchemePop`.

Could you add a one-liner to `MSIDAuthenticationSchemeTest` and `MSIDAuthenticationSchemeSshCertTest` asserting `tokenEndpointParameters` equals `schemeParameters`? Cheap, and it pins the default so a future override can't silently change what those schemes put on the wire.

IdentityCore/src/auth_scheme/MSIDAuthenticationSchemePop.m
antonioalwan: Agreed, removed in 405fb3f. The override now always does `mutableCopy` + `removeObjectForKey:` + `copy`, with no branch on `externalKeyPop`. You are right that the invariant was the real cost and the saved allocation was not worth it at once per token request.

IdentityCore/src/network/request/MSIDTokenRequest.m
antonioalwan: Renamed to `schemeBodyParameters` in 405fb3f. Agreed the old name was misleading, these go into `_parameters` and are form-encoded into the POST body by the request serializer, not set as headers.

IdentityCore/src/auth_scheme/MSIDAuthenticationScheme.m
antonioalwan: Added in 405fb3f:

- `MSIDAuthenticationSchemeTest.testTokenEndpointParameters_whenBearerScheme_shouldEqualSchemeParameters`
- `MSIDAuthenticationSchemeSshCertTest.testTokenEndpointParameters_whenSshCertScheme_shouldEqualSchemeParameters`

Both assert `tokenEndpointParameters` equals `schemeParameters`, which pins the inherited default. The SSH-cert one matters most since that scheme carries a third parameter (`key_id`) that is a genuine eSTS parameter and must keep going on the wire.

`IdentityCoreTests Mac` across all three scheme suites: 28 tests, 0 failures.

josephpab (approved):

